PR #14, opened 2015-12-08
bugfixing & cleaning code 

PidgeyL:


carpedm20:
I think this codes are much cleaner than before. Thx!

Issue #13, opened 2015-12-07
getThreadList seems not to actively check latest posts


PidgeyL:
From some tests I ran, it seems like `getThreadList` does not update when new messages are sent. I had to log in again to get the last messages.
Is it just me, or is this in fact the case?

MikeCroall:
I am finding the same as you.

madsmtm:
Facebook has changed a lot of things, and `fbchat` has been updated, so I'm assuming this is outdated ;)

PR #16, opened 2015-12-15
function read user messages

PidgeyL:


carpedm20:
nice work on `getThreadInfo`!

PidgeyL:
You're welcome :) This fixes #13 only partially, but it seems to be "real time"

PR #21, opened 2015-12-28
Adding the like option to messages

PidgeyL:


carpedm20:
thanks for great update and changing version :+1:

PR #23, opened 2016-01-11
More actions on "listening"

PidgeyL:


Schmavery:
Hey @PidgeyL, I see that you guys support the `typ` event.  Just a heads up that if you also want to get notifications for people typing in a group chat, you'll want to also capture the `ttyp` event, which returns a slightly different object.
As a side note, I'd be curious to know what the `qprimer` event is.
Awesome to see work happening on this repo again.

PidgeyL:
@Schmavery Cool, thanks! I'll put that on my todo list :)
From my understanding (but I have no confirmation), qprimer is something like "Facebook now knows I'm listening". From tests I've ran, I get a qprimer when I start listening, but not when I restart it in the same session.
I'm doing this in my free time though (which is limited currently), so group support might take a while to complete. My next steps are first attachments and indication of blocked messages (Facebooks bot detection).

PR #34, opened 2016-03-24
Remove unnecessary global state

sid-kap:


fbchat/client.py
```diff
@@ -88,11 +88,6 @@ def __init__(self, email, password, debug=True, user_agent=None):
     def _console(self, msg):
         if self.debug: print(msg)
 
-    def _setttstamp(self):
```
sid-kap: This function is only used once (in `login`), so might as well move it into `login`.

fbchat/client.py
```diff
@@ -135,30 +130,24 @@ def login(self):
             self.start_time = now()
             self.uid = int(self._session.cookies['c_user'])
             self.user_channel = "p_" + str(self.uid)
-            self.ttstamp = ''
 
             r = self._get(BaseURL)
             soup = bs(r.text, "lxml")
-            self.fb_dtsg = soup.find("input", {'name':'fb_dtsg'})['value']
```
sid-kap: No need to make `fb_dtsg` a global variable, since it is only used in this function.

fbchat/client.py
```diff
@@ -135,30 +130,24 @@ def login(self):
...
-                'channel' : self.user_channel,
-                'partition' : '-2',
-                'clientid' : self.client_id,
-                'viewer_uid' : self.uid,
-                'uid' : self.uid,
-                'state' : 'active',
-                'format' : 'json',
-                'idle' : 0,
-                'cap' : '8'
-            }
```
sid-kap: `self.form` is never used.

sid-kap:
Got no response, closing for now

Schmavery:
:'(

fbchat/client.py
PidgeyL: I'm not a big fan of this. A function is still a function. Things like this might be needed to be called later on.

sid-kap: Why is this a good reason?

sid-kap: It won't be used in the future. Why would you assume something like that?

PidgeyL: Because it's still a function. It is “clean programming”. 

Also, have you seen the Javascript version this is based on?






From: Sid Kapur
Sent: ‎Wednesday‎, ‎6‎ ‎April‎ ‎2016 ‎22‎:‎41
To: carpedm20/fbchat
Cc: Pidgey





In fbchat/client.py:
> @@ -88,11 +88,6 @@ def __init__(self, email, password, debug=True, user_agent=None):
>      def _console(self, msg):
>          if self.debug: print(msg)
>  
> -    def _setttstamp(self):


Why is this a good reason?

sid-kap: Sorry, my earlier message was vague.

My main concern with this function is that it reads and writes to global state for no reason. It being a separate function is not my biggest concern.

As a secondary concern, however, if you would like to keep it a function, I suggest moving the function inside the `_generatePayload` function (I believe this is what I did in the PR), so that we are not exposing more functions than is necessary. This makes it easier for a newbie to read and understand your code, because they can easily tell that the function is only used inside `_generatePayload`.

PidgeyL: I seem to recall this was also separate in the j JavaScript version (to be verified). If it's not, I don't mind moving the function

-----Original Message-----

sid-kap: It is not a separate function there: https://github.com/Schmavery/facebook-chat-api/blob/master/utils.js#L465

PR #50, opened 2016-07-19
added param for group messages

stanleykylee:
added param to allow for the ability to send messages to group chats using thread_id. if param is not used, defaults to sending to user_id


carpedm20:
@stanleykylee Thanks!

PR #61, opened 2016-08-30
Iss60

### LINKED ISSUE #60: [FIXED] Bug in reception of images
When images without text body is sent, it just called `on_message_error` and not `on_message` making users unable to handle them.

thekindlyone:
fixes #60 by doing a `m['delta'].get('body','')` instead of `m['delta']['body']` , thus avoiding the exception.

Now metadata received in the on_message handler has urls for any image received.


carpedm20:
@thekindlyone Thanks!

PR #68, opened 2016-10-11
Update client.py to post to groups (Issue #63)

aeantipov:
Jut followed the recipe in the issue

PR #74, opened 2016-10-26
other_user_fbid logic was wrong

dcep93:


dcep93:
solves same problem as #68

PR #75, opened 2016-10-26
friend_connect and the like

dcep93:


fbchat/client.py
```diff
@@ -531,6 +545,9 @@ def _parseMessage(self, content):
                         fbid =    m['delta']['messageMetadata']['actorFbId']
                         name =    None
                         self.on_message(mid, fbid, name, message, m)
+                elif m['type'] in ['jewel_requests_add']:
```
dcep93: unclear if this is the only possible friend request type

Issue #86, opened 2016-12-06
Encoding of received messages


vlegoff:
It seems that messages are encoded differently, probably depending on the client being used by the user.  This creates additional struggles for those using non-ASCII languages.  One message can be encoded in Utf-8, which makes the most sense considering the API, I believe.  Another might be encoded differently.  It's difficult to adopt a strategy that can work no matter the encoding being used.  I reviewed last messages received by a client using non-ASCII characters, and couldn't find any "encoding" field in the message model.  Is there some strategy that I could apply in this context?

(Facebook Messenger doesn't seem to mind and display accents just fine each time, no matter the client being used, it seems, so I guess this strategy exists.)

carpedm20:
Maybe now it's work, after e847c551d151e393357a978b98d30e164daa9f94.

JarbasAI:
e847c55. doesnt fix it

simply sending "ç" will cause

123fbid456 said: รง
Exception:
(<type 'exceptions.UnicodeEncodeError'>, UnicodeEncodeError('ascii', u'\u0e23\u0e07', 0, 2, 'ordinal not in range(128)'), <traceback object at 0x7fa780a1bb00>)

Bankde:
Latin, Greek, Thai language should now work correctly.
btw, even though I could generate "ç" to "รง" like you, I couldn't make it crash.

Can you provide more information ? Make sure you have pull the master branch before testing.

carpedm20:
#96 may solved this issue.

madsmtm:
Assuming #96 solved this

PR #96, opened 2017-02-02
Fix fb listening encoding

Bankde:
https://github.com/carpedm20/fbchat/issues/86
According to this issue, I take a look at the code and notice that listening didn't set encoding to 'UTF-8'. Setting to UTF-8 should fix encoding issue.

Testing the latin alphabet and thai alphabet, both seem to work correctly.


carpedm20:
Thanks

PR #97, opened 2017-02-03
Added getAllUsers function

pirelle:


carpedm20:
Thanks for the pull request but what is the use of `self.GENDERS`? I can't see its use in your PR.

pirelle:
I wanted to use it for User model but there was no "gender" field in User model. Also I am not sure where is better place to convert "Gender" value to verbose form but saved convertion table for future :)

carpedm20:
I think we should clean out our codes one day but thanks anyway.

PR #98, opened 2017-02-06
Added functions to save and load session cookies to and from file

madsmtm:
This uses some fairly unknown functions from the "requests" library.

Also, I fixed a spelling mistake

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,22 @@ Example Echobot
     bot.listen()
 
 
+Saving session
+==========================
+
+.. code-block:: python
+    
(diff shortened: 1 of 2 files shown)


Bankde:
Honestly, I don't understand how this works.
It still requires the password from fbchat init function and need to call login function to initialize many variables, isn't it ?

madsmtm:
Yes, it still requires those, and i may have forgot that part. Will update ASAP :)

madsmtm:
@Bankde I've updated the pull request

PR #102, opened 2017-02-06
Added new functions and events. Also fixed sending messages to groups.

madsmtm:
Added functionality to add users in `send`

Fixed sending messages to groups (Sending empty `other_user_fbid` does
not work. It has to be omitted!)

Added new `on_message`, which gets `recipient_id` and `thread_type`.
This will help people sending messages to groups greatly! It also no
longer receives `name`, since this parameter isn't used.

Added event functions `on_people_added` and `on_person_removed`, and
their corresponding calls.


carpedm20:
Thanks for the PRs 👍 I'll review the code before I merge this.

Issue #106, opened 2017-02-07
Problem with EchoBot example


blopa:
Hi!

I was trying to use the EchoBot example, but I'm getting an error:

`(<type 'exceptions.TypeError'>, TypeError('on_message() takes exactly 6 arguments (7 given)',), <traceback object at 0x0000000003A8D088>)`

I just copied and pasted the code from example, login was successfully, this problem happens everytime I got a message.

Thank you

milo123milo:
try to update fbchat
`pip install --upgrade fbchat`

madsmtm:
Yeah, this is my fault, turns out python doesn't work the way i expected (Too much C++, i guess ;) ), when it comes to overriding functions. #108 should fix this

blopa:
Hello guys!

@milo123milo, I tried that, and it says I already have the latest version, v0.7.1.

@madsmtm Thank you. I tried pulling the latest version from this repository, but I'm not sure if this version is overwritting my pip installation or not, because I tried the echobot again and I'm still having the same error.

PR #112, opened 2017-02-12
Fixed args count error when using on_message_new

kienankb:
This should fix the same error addressed by #108  for on_message_new that was brought up for on_message in #106.


kienankb:
Actually hold on this, seems like I managed to break the EchoBot example.  Let me investigate and fix.

kienankb:
Nevermind, it's not breaking, I just forgot to update the function signature in my local copy of the example by removing the author_name argument.  All good!

PR #115, opened 2017-03-07
Query ignore fix

botcs:
- Refactored `getThreadList` parameter `end` which is ignored by the `_post` method if `end > 20`.
- Refactored `getThreadInfo` parameter `start` which is ignored, messages are always returned from the last message with 0 index.
- Renamed `getThreadInfo` parameter `end` to `length` which better represents the behaviour
- updated `getUserInfo` to accept both:
  - string IDs starting with `fbid:`  produced by entries of `messages = client.getThreadInfo(...)`
  - string IDs produced by `client.getUsers('<friend name>')[0].uid` 

assertion error is thrown when parameters are used improperly

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -521,7 +521,7 @@ class Client(object):
         :param thread_type: (optional) change from 'user' for group threads
         """
         
-        assert(length > 0, 'length must be positive integer, got %d'%length)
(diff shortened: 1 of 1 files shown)


carpedm20:
Thanks for the PR!

PR #122, opened 2017-04-04
added the hability to control fbchat from an external event loop

thgcode:
I needed to implement this feature for an application I'm developing, and thought that this could be useful for other people too.


carpedm20:
Thanks! This will be definitely helpful.

PR #129, opened 2017-05-01
fix inconsistencies, remove some deprecated and unused things

Dainius14:
just a small general clean up

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -27,7 +27,7 @@ import sys
 
 # Python 3 does not have raw_input, whereas Python 2 has and it's more secure
 try:
-    in put = raw_input
+    input = raw_input
 except NameError:
     pass
 
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -59,7 +59,7 @@ class Client(object):
     documentation for the API.
     """
 
-    def __init__(self, email, password, debug=True, user_agent=None, max_retries=5, do_login=True):
+    def __init__(self, email, password, debug=True, info_log=True, user_agent=None, max_retries=5, do_login=True):
         """A client for the Facebook Chat (Messenger).
 
         :param email: Facebook `email` or `id` or `phone number`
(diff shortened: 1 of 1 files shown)


Bankde:
Hi, looks great :)

I would like to add a point this one.
```
-        s = raw_input('Please enter your 2FA code --> ')
+        s = input('Please enter your 2FA code --> ')
```
raw_input does not exist in python3, how does no one ever raise issue on this one ? O_O
I suggest adding this at the top after import. So the code can be securely compatible to both py2 and py3.
```
try:
   input = raw_input
except NameError:
   pass
```
In python2.7, we need to use raw_input because of security issue.  For personal use, might be ok but it becomes serious issue when someone decide to use this lib as open service.

Dainius14:
Hey. I've fixed input for P2.
In addition I've set up info level logging, where only logging in and message sending is logged, debug is left for actual content sent/received. I've also added default recipient, to make it easier to send to the same person. 
A probably breaking change would be that I've changed `send` signature, so now it accepts boolean `is_user` instead of `message_type` where it can have only two values.

On a separate note, it'd be great to make some consistency across method names, return values and overall project, since now they names are both camel cased and separated with underscores, methods some return bool, some none.
With client getting over 1k lines it'd be not bad to start separating file into smaller pieces. Who's up for that?

Why? I need all of this for my chat bot. Could have done this months ago, but forgot and abandoned project. Since, now I'm doing it again all this extra stuff given back to community would be not too bad.

Bankde:
Hi, sry for being unclear.
You need to put
```
+        try:
+            input = raw_input
+        except NameError:
+            pass
```
at the top, after importing module. Otherwise python will interpret input (that is declared within the function) as variable and throw unbound exception.

IMO, unit test is necessary for this project right now before proceeding to something bigger.

Dainius14:
I've changed that

Bankde:
```
+    in put = raw_input
```
typo ?
Everything else looks great and clean. Thank you :)

Dainius14:
I don't even understand how that could have happened...

carpedm20:
Thanks @Dainius14 for the PR and @Bankde for the review! It's become more cleaner and better now. I'll merge this. Thx!

PR #131, opened 2017-05-05
Fix for getUserInfo method throwing an error

tmcadam:
The participants property on thread object is returning a list of unicodes. If you then pass these unicodes to getUserInfo, they are not being picked up in the if statement on line 905 in the fbidStrip method, so the fbidStrip method was returning a tuple of None objects. The conditional now matches and reformats both str and unicode objects.


carpedm20:
I see. Thanks for the PR!

PR #134, opened 2017-05-07
add event hooks, change method names

Dainius14:
This would be a **not** backwards compatible version and is still a little WIP.

Things changed:
* Better event system. Now it is possible to simply add your own event handlers via `=+` and `=-`  operators, no need to overwrite any methods. Subscribing this way is IMO a cleaner way, and you can have multiple different handlers.
* Unification of method names. Now all methods are camelCased, unlike before where some were underscore separated and some not.
* Reworked events. This is still WIP, but unified parameter position as much as possible and added some extra ones. IMO all of them need to return an `message_id` and a `timestamp`, for future-proof.
* Added a few extra events. `onTitleChange` and `onColorChange`. Going to add a few more from my old source.
* Changed stickers location. Removed `stickers.py`, moved them to `models.py` and made as an enum. This should be little more logical than what we had now, a whole separate file for literally 6 lines of code.

I hope I didn't forget anything. Would appreciate any input.


fbchat/client.py
```diff
@@ -794,71 +771,120 @@ def _parseMessage(self, content):
...
+
+                    # Title change
+                    elif delta_type == "change_thread_nickname":
+                        mid = metadata["messageId"]
+                        author_id = metadata["actorFbId"]
+                        changed_for = delta["untypedData"]["participant_id"]
+                        new_title = delta["untypedData"]["nickanme"]
+                        thread_id = metadata["threadKey"].get("threadFbId")
+                        ts = metadata["timestamp"]
+                        self.onTitleChange(mid=mid, author_id=author_id, changed_for=changed_for, new_title=new_title,
```
madsmtm: Though this will always have a `msg_type` of `MsgType.GROUP`, I would still add this as a parameter to the event

Dainius14: Actually this is quite interesting as this event would get fired for both thread and nickname changes. 

Maybe it would be even better to separate events to group and user events? For example `onMessage` and `onGroupMessage`, `onTyping`, `onGroupTyping` and so on. It'd be a little cleaner - although this is subjective.

I will update README.md once I get all the things right.

Having backwards compatibility *would* be possible. But I don't think it would be realistic to keep it forever, too many duplicate things to maintain. This update would have a major version bump and maybe some kind of warning when installing through pip.

madsmtm: About the `onMessage` vs `onGroupMessage` and `onChatMessage`: I would vote for a single `onMessage`, since it would require `sendToChat()` and `sendToGroup()` functions, which in turn would mean a lot of duplicate code

madsmtm: And yeah, with a project like this one, which relies on the requests Facebook makes, backwards compatibility is rarely possible anyways

Dainius14: `sendToGroup` and `sendToPerson` could be easily solved by having one main private send method (like now) and other methods which only call `send` with correct parameters.\

That way we could have 
`onPrivateMessage(mid, author_id, message, ts, metadata)`
`onGroupMessage(mid, author_id, message, thread_id, ts, metadata)`
`sendPrivateMessage(message, recipient_id)`
`sendGroupMessage(message, thread_id)`

and so on.

You can see on my old fork here https://github.com/Dainius14/fb-chat-bot/blob/master/fbchat/client.py#L313 how it would look more or less.

madsmtm: Yeah, I agree, it's possible, but it's rare that when implementing some sort of bot, that you'd want to act differently on groups than on private chats. That's what I mean by duplicate code, not so much duplicate code in `fbchat`, but duplicate code in user code. Apart from that, functions like `send`, `sendRemoteImage` and so on should also take a `MsgType` as a parameter instead of `is_user`

madsmtm: If the "Facebook API" was designed in a different way, with private conversations and group conversations being split, `fbchat` should probably be designed with that in mind, but since they're not split, this is probably the best approach

Dainius14: Yeah, logical. 

Though some sending methods could be separated, such as sending like, adding participants. You can't send both a like and a message in one request. Old clients show both the message and a sticker, but new clients only show a sticker. And also it's not possible to add a user and a send a message simultaneously.

madsmtm: Yup, that's definitely something that needs to be done, making a `_send` function or something, and then, as you said, `sendLike` and `sendMessage`

madsmtm: I would actually recommend `sendSticker(self, size, emoji=None)`, and then have it send the default sticker when `emoji` is `None`

madsmtm: Specifically, if you need to send a sticker, it goes something like this:

```python
data['body'] = emoji
data['tags[0]'] = 'hot_emoji_size:' + size # Size must be one of 'large', 'medium' and 'small'
data['tags[1]'] = 'hot_emoji_source:picker'
```

madsmtm:
Seems good to me! Very good work!!!

madsmtm:
Hey @Dainius14, I've added a development branch, maybe you could make a new PR with your latest progress, and I'll add it to the branch :)

PR #140, opened 2017-05-10
Updated send methods

Dainius14:
Instead of exposing one big `send()` method, I made it private and created other smaller methods with specific purpose and more understandable parameters:
* `sendMessage()`
* `sendRemoteImage()`
* `sendLocalImage()` - doesn't work for me, response payload is empty
* `sendEmoji()`
* `addUsersToChat()`
* `removeUserToChat()`
* `getThreadInfo()`

Updated tests according to new methods. Test it out on your machine first


madsmtm:
I've updated the development branch now. One thing is that you made all the parameters type specific, which is python 3.x only, sadly. I had to remove that, since we should keep supporting at least both python 2.7 and python 3.6. Other than that, good work for now :)

PR #141, opened 2017-05-10
reintroduce things skipped on conflict

Dainius14:
Size should be EmojiSize so not to cause confusion

IMO we should ditch all deprecation warnings, make it backwards-incompatible and bump to version 1.0 once I add more chat methods. there is no correct way to keep on_* methods with a warning, they would be called everytime and generate the warning everytime for everyone. this is like dragging a dead horse around.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *py[co]
 
+.idea/
+
 # Test scripts
 *.sh
 
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -20,6 +20,7 @@ from bs4 import BeautifulSoup as bs
 from mimetypes import guess_type
 from .utils import *
 from .models import *
+from .event_hook import *
 import time
 import sys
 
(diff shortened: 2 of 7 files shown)


madsmtm:
You're right about the `EmojiSize`. About the deprecation warnings, it's actually possible using `hasattr(type(self), 'on_*')` in `__init__`.

Sketch of how it might work (not tested)
```python
if hasattr(type(self, 'on_message')):
    deprecation('Client.on_message', deprecated_in='0.10.2', details='Use new event system (Client.onMessage) instead')
    self.onMessage += lambda mid, author_id, message, thread_id, thread_type, ts, metadata:\
    self.on_message(self, mid, author_id, None, message, metadata)
```

And though I'm not keep on having this excess code and all that, I've realised that there's actually a lot of people using `fbchat`. Besides, we don't have to keep backwards compatibility forever, the `_console` function can probably be removed soon. Also, I don't know of a way to warn people through `pip`, so we're basically going to screw a lot of people over if we don't do this right ;)

PR #145, opened 2017-05-16
Add more events and send methods, some fixes

Dainius14:
Will add several more I think and then it will be good to go

On a separate note, on every deprecation message we need to write something "will be removed in version 1.3" or smth. It's bad to keep a spaghetti of old and new code in the same plate.

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -264,8 +264,8 @@ class Client(object):
         self.payloadDefault = {}
         self.client_id = hex(int(random()*2147483648))[2:]
         self.start_time = now()
-        self.uid = int(self._session.cookies['c_user'])
-        self.user_channel = "p_" + str(self.uid)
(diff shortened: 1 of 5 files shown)


madsmtm:
Nice work!

IMO uid's should be integers, and not strings.

fbchat/client.py
```diff
@@ -456,13 +446,29 @@ def resetDefaultThread(self):
         self.default_thread_id = None
         self.default_thread_type = None
 
+    def _setThread(self, given_thread_id, given_thread_type):
```
madsmtm: Maybe rename this to `_getThread`?

Dainius14:
Well I think in this way - are they manipulated in a mathematical way? We never need to see if one ID should be bigger or smaller. So they should be strings. Furthermore, FB  *mostly* sends them as strings, not ints.

PR #148, opened 2017-05-29
Opinion on new version needed

madsmtm:
DON'T MERGE THIS!

I've been working with @Dainius14 the past month on the `development` branch, and we've arrived at this result. There are still a lot of things missing, but I need your input on these matters. I'll start by outlining what's changed:

- Added a [Sphinx](http://www.sphinx-doc.org/) powered documentation, which is unlisted, but live on [Read the Docs](https://fbchat.readthedocs.io)
- Deprecated a lot of functions, including all the old `on_x` methods, and added a whole new range of events
- Added more functions for interacting with threads, and streamlined most of the input arguments (all thread type inputs are now called `thread_type`, and so on)
- Added new event system, which works by adding events like so:
```python

def onMessageFunc(...):
    pass

client.onMessage += onMessageFunc
```

Some of the important things that are still missing are:

- A better way to fetch information. Currently, all the `getX` methods have weird and non-standardised return values
- Documentation of all event methods
- The documentation is still missing a few things, and it's probably not very concise in a few places

**Now to the important part!**

I need your opinion on three things: Whether we should keep supporting v. 0.9.x, whether we should introduce the new event system, and which python versions we should be supporting

1. I've tried to make this version be backwards compatible with v. 0.9.x, though a lot of functionality is probably still broken. This is not impossible to fix, and not impossible to maintain, though it may not really be worth it. Maybe we should go to v. 1.0.x, and say that it's incompatible with previous versions, and then work on proper deprecations and such in all future versions? And then people can just downgrade to 0.9.x if they need to?

2. @Dainius14 introduced the new event system, and though I was a huge fan at the start, I've found that the old way of doing it is both easier to explain, easier to use and easier to document (Currently I haven't found a good way of documenting the new events). I might use the event script in my own projects, but I think it's something that's up to the user to implement, and not something we should be doing

3. I've tried to keep support for python 2, and I've installed and tested `fbchat` against all Python versions that are still receiving updates. I think we should keep it that way, keep supporting those versions of python, though if you have an opinion on this, please let me know

I hope I've been clear on what I meant, and I hope you'll help out by leaving your opinion :)


madsmtm:
This PR has been open for 10 days with still no comments, so I'm going to wait a few more days, and then I'll implement it into the master branch

Bankde:
Everything here is my very own opinion.
It is not necessary to always keep things compatible. I would say go ahead, start new 1.0.x and delete all deprecated functions. Better APIs are better future. Better to move on; but making sure that new APIs are stable.

However, put old version 0.9.x in "releases" so people can still download it as they want.

For old version, you can keep them up-to-date to current facebook APIs as you wish but not necessary. This is not commercial and not even official APIs from facebook. People need to understand and adapt.

Thank you for your hard works :)

Dainius14:
I was rooting for the drop of support for old API from the very beginning when I started changing it. It's the only way to go.

madsmtm:
@Bankde Thanks for your opinion, and though I didn't at first, I agree with you. I'll wait a few more days, and then I'll get back to merging it. Also, I probably won't be supporting the previous versions ;)

@Dainius14 You were right ;). Though I do think deprecating things in versions after this one should be done where feasible

PR #155, opened 2017-06-05
Logging handler fix

DennyWeinberg:
Do not add each time a new StreamHandler to the global logger object
when you use a new Client in the same runtime/import


madsmtm:
This is fixed in on the `development` branch, which will be merged pretty soon ;)

PR #169, opened 2017-06-26
Handle empty participant_customizations field

enwar3:
The participant_customizations field is sometimes an empty array

### DIFF
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -44,7 +44,7 @@ def get_customization_info(thread):
         rtn['nicknames'] = {}
         for k in info['participant_customizations']:
             rtn['nicknames'][k['participant_id']] = k.get('nickname')
-    elif info['participant_customizations']:
+    elif info.get('participant_customizations'):
         _id = thread.get('thread_key', {}).get('other_user_id') or thread.get('id')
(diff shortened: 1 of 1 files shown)


fbchat/graphql.py
```diff
@@ -44,7 +44,7 @@ def get_customization_info(thread):
         rtn['nicknames'] = {}
         for k in info['participant_customizations']:
             rtn['nicknames'][k['participant_id']] = k.get('nickname')
-    else:
+    elif info['participant_customizations']:
```
madsmtm: You are right in this change! But if `info['participant_customizations']` is not set, this will still throw an error. Maybe we should have it be `info.get('participant_customizations')`?

enwar3: Yea so far I've only seen it come back as an empty array, I haven't seen the field not set yet, so I only wanted to handle that and didn't want to make any additional assumptions. But down for whatever you think is better!

madsmtm: Maybe it's not really *that* important here, but in general, I think we should be as prepared as possible  ;)

enwar3: Gotcha! Just wasn't sure what our policy was on throwing exceptions. Changed to ```.get()```

PR #174, opened 2017-07-05
No FileNotFoundError in py2

Bankde:
In setup.py
There is no FileNotFoundError in Python2
However, since we don't do anything with exception anyway, using IOError is clean and work for both py2 and py3.


madsmtm:
Oh right, thanks for the fix!

PR #190, opened 2017-08-10
Improve login speed and reliability

madsmtm:
This should increase login speed and reliability, by making one less request when logging in, and adding an extra check if request contains the parameter `next`

I've also added some custom exceptions, since it'll be useful to see where the exception is actually coming from.

This has not yet been tested with 2FA, and while it should work, I need someone to test this


madsmtm:
Wayyy outdated, I can't remember what I did anymore

PR #197, opened 2017-08-30
Seen all message from group chat

tdtgit:


madsmtm:
I'm sorry, but I don't really see the point in demonstrating this. This is not a required step to make a "keepbot". Perhaps you could elaborate on your reasoning, and maybe include some comments where you feel necessary. Other than that, I'd like to thank you for trying to improve the documentation, it's a long way from finished ;)

tdtgit:
This PR make the bot more real to the user, he will "seen" all when have new message and make me feel I'm not alone when programming at night, jdk. By the way your bot is awesome, me and my friends have a lot of fun time will KeepBot :P

madsmtm:
It's nice to see that you're enjoying the bot ;)
But adding this is probably a bad idea, as it makes it seem like it's required to maintain the basic functionality, which it is not.
If you feel like an example of how `markAsDelivered` and `markAsRead` is used is needed, make a new one instead :)
Besides, both `markAsDelivered` and `markAsRead` are currently undocumented, and unreliable. If possible, It'd be nice if this was fixed before anything else! ;)

PR #210, opened 2017-10-02
Add Gif support to send(Local/Remote)Image

WeiTang114:
sendImage(), sendLocalImage(), and sendRemoteImage() support different data types, but some keys need to change for GIF specially.


WeiTang114:
I'm bad at doing PR. 
Sorry to include too many commits.

Issue #177, opened 2017-07-07
Accept request message for new contact


3dluis:
There is no method for accepting new contacts when writing to you for the first time.

kapi2289:
Added in #318 `moveThreads` method. When you move pending thread to inbox it'll be accepted!

madsmtm:
Fixed in release `v1.4.0`

PR #211, opened 2017-10-02
Enable fetching pending/archived threads

WeiTang114:
Add "type" parameter to fetchThreadList().
type can be 'inbox', 'pending' or 'archived'

If set to 'pending', it can fetch messages from unknown users.
It is quite useful to build a service accepting requests from anyone.

For example, in doOneListen(), fetch pending messages once for a while
to handle the messages from strangers.

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -748,11 +748,12 @@ class Client(object):
 
         return list(reversed([graphql_to_message(message) for message in j['message_thread']['messages']['nodes']]))
 
-    def fetchThreadList(self, offset=0, limit=20):
+    def fetchThreadList(self, offset=0, limit=20, type='inbox'):
         """Get thread list of your facebook account
 
         :param offset: The offset, from where in the list to recieve threads from
         :param limit: Max. number of threads to retrieve. Capped at 20
+        :param type: (optional) "inbox", "pending", "archived"
         :type offset: int
         :type limit: int
         :return: :class:`models.Thread` objects
@@ -763,10 +764,16 @@ class Client(object):
         if limit > 20 or limit < 1:
             raise FBchatUserError('`limit` should be between 1 and 20')
 
+        if type in ['inbox', 'pending', 'archived']:
+            if type == 'archived':
+                type = 'action:archived'
+        else:
+            raise ValueError('thread_type must be "inbox", "pending" or "archived"')
+
         data = {
             'client' : self.client,
-            'inbox[offset]' : offset,
-            'inbox[limit]' : limit,
+            type + '[offset]' : offset,
+            type + '[limit]' : limit,
         }
 
         j = self._post(self.req_url.THREADS, data, fix_request=True, as_json=True)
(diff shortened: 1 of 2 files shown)


fbchat/client.py
```diff
@@ -748,11 +748,12 @@ def fetchThreadMessages(self, thread_id=None, limit=20, before=None):
 
         return list(reversed([graphql_to_message(message) for message in j['message_thread']['messages']['nodes']]))
 
-    def fetchThreadList(self, offset=0, limit=20):
+    def fetchThreadList(self, offset=0, limit=20, type='inbox'):
```
madsmtm: Overriding the built in function [type](https://docs.python.org/library/functions.html#type) is bad practice. Perhaps we could rename it to something else?

fbchat/client.py
```diff
@@ -763,36 +764,44 @@ def fetchThreadList(self, offset=0, limit=20):
         if limit > 20 or limit < 1:
             raise FBchatUserError('`limit` should be between 1 and 20')
 
+        if type in ['inbox', 'pending', 'archived']:
+            if type == 'archived':
+                type = 'action:archived'
+        else:
+            raise ValueError('thread_type must be "inbox", "pending" or "archived"')
```
madsmtm: We're using custom exceptions, so instead of `ValueError`, it's `FBchatUserError`. This was a recent change, and I'm still not sure it's a good idea, perhaps we should just drop that, and use `ValueError` instead?

fbchat/client.py
```diff
@@ -748,11 +748,12 @@ def fetchThreadMessages(self, thread_id=None, limit=20, before=None):
 
         return list(reversed([graphql_to_message(message) for message in j['message_thread']['messages']['nodes']]))
 
-    def fetchThreadList(self, offset=0, limit=20):
+    def fetchThreadList(self, offset=0, limit=20, type='inbox'):
         """Get thread list of your facebook account
 
         :param offset: The offset, from where in the list to recieve threads from
         :param limit: Max. number of threads to retrieve. Capped at 20
+        :param type: (optional) "inbox", "pending", "archived"
```
madsmtm: Following the pratice in `fbchat`, perhaps we should use an `Enum` instead of string values? This would allow it to be reused across different functions, and it looks neat and clean ;)

madsmtm:
Also, would the `pending` parameter be the proper way to address #177?

WeiTang114:
@madsmtm 
Yes it is.
My implementation is [here](https://github.com/WeiTang114/BOOSTicker/blob/master/stickerbot.py) (at line 90).
I fetch and see if there are any pending threads every PENDING_THREAD_DURATION_SEC seconds.

fbchat/client.py
WeiTang114: oh this is bad. I'll figure out another good name :)

fbchat/client.py
WeiTang114: OK, Thanks!

fbchat/client.py
WeiTang114: I'm not sure either :sweat_smile: 
I think I'll follow the FBchatUserError way for now. Thanks!

madsmtm:
Thank you for your contribution, and thanks for the example showing how to accept pending messages! ;)

maraid:
@WeiTang114 thank you, big help right now. 

For others who stumble upon this, it might be a good idea to use onInbox callback to detect messages in the pending box rather than reading it every X seconds (well depends on how often you are expecting messages anywhere). I just piped it to onMessage where i handle all normal messages.
```python
def onInbox(self, unseen=None, unread=None, recent_unread=None, msg=None):
    thread = self.fetchThreadList(self, limit=1, thread_location=models.ThreadLocation.PENDING)[0]
    msg = self.fetchThreadMessages(thread_id=thread.uid, limit=1)[0]
    if not msg.is_read:
        self.onMessage(author_id=msg.author, message_object=msg, thread_id=thread.uid, thread_type=thread.type)
```

PR #212, opened 2017-10-02
Add Gif support to send(Local/Remote)Image

WeiTang114:
(Fix the last PR #210)
sendImage(), sendLocalImage(), and sendRemoteImage() support different data types, but some keys need to change for GIF specially.

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -962,6 +962,7 @@ class Client(object):
         :param message: Additional message
         :param thread_id: User/Group ID to send to. See :ref:`intro_threads`
         :param thread_type: See :ref:`intro_threads`
+        :param is_gif: if sending GIF, True, else False
         :type thread_type: models.ThreadType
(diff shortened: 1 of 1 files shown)


fbchat/client.py
```diff
@@ -938,7 +938,7 @@ def sendEmoji(self, emoji=None, size=EmojiSize.SMALL, thread_id=None, thread_typ
 
         return self._doSendRequest(data)
 
-    def _uploadImage(self, image_path, data, mimetype):
+    def _uploadImage(self, image_path, data, mimetype, is_gif=False):
```
madsmtm: In here, we are getting the `mimetype` of the image, perhaps we can use that to differentiate between gifs/other images instead?

WeiTang114: Thanks. I tried that and it works! Already changed to that.

WeiTang114:
Update thanks to @madsmtm 's feedback.

PR #214, opened 2017-10-06
new feature eventReminder added

svaikstude:
Added new feature to send an event reminder to a thread. It only creates a new event, but does not update an old one, if there already is an event under the same name.


madsmtm:
This doesn't work in python2.7, but I've merged it anyways, because I don't have an easy (nor a hard) fix

PR #218, opened 2017-10-14
Added initial integration for messenger rooms

ekohilas:
Updated for compatibility with python2.7


fbchat/graphql.py
```diff
@@ -118,6 +118,26 @@ def graphql_to_group(group):
         message_count=group.get('messages_count')
     )
 
+def graphql_to_room(room):
+    if room.get('image') is None:
+        room['image'] = {}
+    c_info = get_customization_info(room)
+    return room(
```
madsmtm: `Room`, not `room` ;)

ekohilas: How are you testing?
It didn't appear when I tested this :(

PR #238, opened 2017-12-27
Fix attachment parsing for newer structure

lobstr:
Message attachments are structured differently now (no more 'attach_type'), and shared items come as 'extensible_attachment'.


madsmtm:
Thanks for this, these are some good corrections! I went ahead and added the onTyping method, and documented it in the process.

PR #254, opened 2018-01-30
modify AudioAttachment function

sdnian:
Get audio message information for fetchThreadMessages()

### DIFF
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -112,8 +112,9 @@ def graphql_to_attachment(a):
     elif _type == 'MessageAudio':
         return AudioAttachment(
             filename=a.get('filename'),
-            playable_url=a.get('playable_url'),
-            duration=a.get('playable_duration_in_ms')
+            url=a.get('playable_url'),
+            duration=a.get('playable_duration_in_ms'),
+            audio_type=a.get('audio_type')
         )
     elif _type == 'MessageFile':
         return FileAttachment(
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -109,6 +109,12 @@ def graphql_to_attachment(a):
             large_image=a.get('large_image'),
             uid=a.get('legacy_attachment_id')
         )
+    elif _type == 'MessageAudio':
+        return AudioAttachment(
+            filename=a.get('filename'),
+            playable_url=a.get('playable_url'),
+            duration=a.get('playable_duration_in_ms')
+        )
     elif _type == 'MessageFile':
         return FileAttachment(
             url=a.get('url'),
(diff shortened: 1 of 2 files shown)


fbchat/models.py
```diff
@@ -267,10 +267,20 @@ def __init__(self, url=None, size=None, name=None, is_malicious=None, **kwargs):
         self.name = name
         self.is_malicious = is_malicious
 
-class AudioAttachment(FileAttachment):
-    def __init__(self, **kwargs):
-        """Represents an audio file that has been sent as a Facebook attachment - *Currently Incomplete!*"""
-        super(StickerAttachment, self).__init__(**kwargs)
+class AudioAttachment(Attachment):
+    # Name of the file
```
madsmtm: These comments should start with a colon, like this: `#: comment`, so they'll show up in the sphinx autogenerated docs on [ReadTheDocs](http://fbchat.readthedocs.io/)

fbchat/graphql.py
```diff
@@ -109,6 +109,12 @@ def graphql_to_attachment(a):
             large_image=a.get('large_image'),
             uid=a.get('legacy_attachment_id')
         )
+    elif _type == 'MessageAudio':
+        return AudioAttachment(
```
madsmtm: The `a` dict contains an `audio_type`, maybe we should include this in the `AudioAttachment` too?

fbchat/models.py
```diff
@@ -267,10 +267,20 @@ def __init__(self, url=None, size=None, name=None, is_malicious=None, **kwargs):
...
+    filename = None
+    # playable_url
+    playable_url = None
+    # duration
```
madsmtm: Maybe you could elaborate on the documentation comments? Something like `Duration of the audioclip in milliseconds`

fbchat/models.py
```diff
@@ -267,10 +267,20 @@ def __init__(self, url=None, size=None, name=None, is_malicious=None, **kwargs):
...
 
-class AudioAttachment(FileAttachment):
-    def __init__(self, **kwargs):
-        """Represents an audio file that has been sent as a Facebook attachment - *Currently Incomplete!*"""
-        super(StickerAttachment, self).__init__(**kwargs)
+class AudioAttachment(Attachment):
+    # Name of the file
+    filename = None
+    # playable_url
+    playable_url = None
```
madsmtm: Just using `url` should be sufficiently describing here ;)

sdnian:
Thank you for your reply. I've follow your suggestion to modify these codes. But I don't know how to update? Sorry, this is my first time to do this. And I am not a programmer.

madsmtm:
Think you updated it properly, it looks good now!

PR #258, opened 2018-02-16
Change logging mode from DEBUG to INFO

Dante383:
Debug mode caused many trash to get printed in the console, and passing logging level in class initiation didn't seem to fix it.

### DIFF
--- a/fbchat/utils.py
+++ b/fbchat/utils.py
@@ -17,7 +17,7 @@ except NameError:
 
 # Log settings
 log = logging.getLogger("client")
-log.setLevel(logging.DEBUG)
+log.setLevel(logging.INFO)
 # Creates the console handler
 handler = logging.StreamHandler()
 log.addHandler(handler)


Dante383:
Maybe it's because the library uses log.info everywhere, and doesn't sort messages for info, debug, error, etc. This is a work-around, temporary solution.

DeltaF1:
I've had success setting the logging level higher in the initializer, are you sure it's not working? This line seems to show that it does:
https://github.com/carpedm20/fbchat/blob/314252480985827c4ebdb6c1ad2b87a7bb2d2a94/fbchat/client.py#L69

madsmtm:
The real solution is to not use the log handlers at all, but this would probably be backwards incompatible. The temporary solution until we break backwards compatability is to just do what @DeltaF1 suggested.

brianmatute011:
Hey alguien acá

brianmatute011:
Tengo un problema

brianmatute011:
Al ejecutar

madsmtm:
Hi @brianmatute011, I sadly don't speak Spanish, can you try to elaborate on your issue in English?

If you're not a native speaker, you can use a service like [Google Translate](https://translate.google.com/) or similar to help you out 😊

brianmatute011:
sorry not knew that not dominated the spanish the point is q and treaty of
carry fbchat to the terminal of my smartphone ... bone with termux nesecito
have fbchat in all parties with me but for the q i see not is possible do
believes q me can help ...

brianmatute011:
is q i am very busy lately ... barely i check my e-mail ... sometimes alone
for issues private important but the greater percentage of my time my
connection to internet is in other domain higher and the gmail not me is
useful in in that moment ... but nevertheless sorry the backwardness ... in
the answer

madsmtm:
I'm sorry @brianmatute011, I really don't understand what you're trying to say. You said something about "having fbchat in all parties is not possible"? Are you having problems with listening and `onMessage`? Or are you having problems with retrieving threads/parties?

brianmatute011:
the that treatment of tell is that fbchat me to been of much help and not
and had success in the installation of this tool in my smartphone ¿could
help me?

madsmtm:
As far as I understand, you're having trouble installing `fbchat` on your smartphone, using [termux](https://termux.com/)? I haven't tried it myself, so I probably can't help you very much. But did you manage to install Python?

brianmatute011:
yes and all its bookstores ... but it fails in the execution ... could you
work based on that? ... there are people who not only install their tools
on their laptop and that is my case ... I need to acquire a specific
installation on my smartphone ... because it has had great results in my
work performance

madsmtm:
Just to clarify, you're having trouble _installing_ `fbchat` on your phone?

madsmtm:
Fixed in d3a0ffc47802f6593ff8871795f53cf21bda714e

PR #266, opened 2018-02-22
Add last_message_timestamp support

### LINKED ISSUE #264: Thread.last_message_timestamp attribute seems empty all the time
Hi,

The attribute 'last_message_timestamp' of Thread object seems empty.
It's a bug or the user was supposed to call a function to fill this value ?
It would be nice if fetchThreadList() take care of filling the value.

It becomes really useful to fill 'before' parameter of fetchThreadList().

After that, it would be easy to provide a meta function to fetch all threads (bypassing FB limit of 20 threads per request)
(I will PR as soon as I find time ;) )

Bye

2FWAH:
Add last_message_timestamp support for one to one and group conversations.

Close #264

### DIFF
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -181,6 +181,7 @@ def graphql_to_thread(thread):
         c_info = get_customization_info(thread)
         participants = [node['messaging_actor'] for node in thread['all_participants']['nodes']]
         user = next(p for p in participants if p['id'] == thread['thread_key']['other_user_id'])
+        last_message_timestamp = thread['last_message']['nodes'][0]['timestamp_precise']
 
         return User(
             user['id'],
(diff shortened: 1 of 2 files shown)


carpedm20:
Thanks for the PR.

fbchat/graphql.py
```diff
@@ -205,6 +207,7 @@ def graphql_to_group(group):
     if group.get('image') is None:
         group['image'] = {}
     c_info = get_customization_info(group)
+    last_message_timestamp = group['last_message']['nodes'][0]['timestamp_precise']
```
madsmtm: This line fails when calling `searchForGroups`, with a `KeyError: u'last_message'`. I've changed it to
```
if 'last_message' in group:
    last_message_timestamp = group['last_message']['nodes'][0]['timestamp_precise']
```

madsmtm:
I'm sorry about being late on this, I've been sick for the last week.

PR #269, opened 2018-03-18
Fix markAsRead and fetchUnread; fixes #261

### LINKED ISSUE #261: how to get unread threads and mark messages as read
hi, i just want to know how to mark messages as read. i tried the markAsRead and markAsSeen but when i check on the facebook messenger it is still flagged and unread/unseen. thanks.

and how to get unread threads.

thanks

gave92:
This fixes the `markAsRead` and `fetchUnread` functions.
The problem with `markAsRead` was that python encodes boolean values in requests as "True/False" while Facebook expects lowercase "true/false".


fbchat/client.py
```diff
@@ -20,6 +20,8 @@ class Client(object):
     See https://fbchat.readthedocs.io for complete documentation of the API.
     """
 
+    ssl_verify = True
```
madsmtm: I like the extra customization you sneaked in here ;), though I don't think we should say "debugging with Fiddler", just something like "debugging with a proxy"

fbchat/client.py
```diff
@@ -797,18 +799,26 @@ def fetchUnread(self):
         :raises: FBchatException if request failed
```
madsmtm: `fetchUnread` seems to work now, but it's still not documented, perhaps you could write something in there?

fbchat/client.py
```diff
@@ -797,18 +799,26 @@ def fetchUnread(self):
...
-        j = self._post(self.req_url.THREAD_SYNC, form, fix_request=True, as_json=True)
+        j = self._post(self.req_url.UNREAD_THREADS, form, fix_request=True, as_json=True)
 
-        return {
-            "message_counts": j['payload']['message_counts'],
-            "unseen_threads": j['payload']['unseen_thread_ids']
-        }
+        return j['payload']['unread_thread_fbids'][0]['other_user_fbids']
+
+    def fetchUnseen(self):
```
madsmtm: Can't verify whether `fetchUnseen` works or not, since I don't really know what it does. Add some documentation ;)

fbchat/client.py
```diff
@@ -1230,28 +1240,28 @@ def setTypingStatus(self, status, thread_id=None, thread_type=None):
...
         return r.ok
 
-    def markAsRead(self, userID):
+    def markAsRead(self, threadID):
         """
         .. todo::
             Documenting this
         """
         data = {
+            "ids[%s]" % threadID: True,
```
madsmtm: You said the issue was the capitilasation of the `True`, and I can verify that, but if that's the only issue, why not just write `'true'`?

fbchat/utils.py
```diff
@@ -225,3 +231,35 @@ def get_emojisize_from_tags(tags):
         except (KeyError, IndexError):
             log.exception('Could not determine emoji size from {} - {}'.format(tags, tmp))
     return None
+
+def encode_params(data):
```
madsmtm: I'm really not a fan of including something like this. I realize that it solves the problem, but as a first, it doesn't work in Python 3, and second, it adds a lot more complexity. As I said, just using `'true'` would solve the problem too. If this doesn't do anything else, I'd suggest removing it

fbchat/client.py
```diff
@@ -137,11 +139,11 @@ def _graphql(self, payload, error_retries=3):
             raise e
 
     def _cleanGet(self, url, query=None, timeout=30):
-        return self._session.get(url, headers=self._header, params=query, timeout=timeout)
+        return self._session.get(url, headers=self._header, params=encode_params(query), timeout=timeout, verify=self.ssl_verify)
```
madsmtm: You forgot to add the `verify` in `_postFile`

fbchat/client.py
gave92: Done

fbchat/client.py
gave92: Done

fbchat/client.py
gave92: Done

fbchat/client.py
gave92: fetchUnseen should return a list of "new" threads but I was not able to make FB return anything except an empty array :)
That request really exists though (seen in fiddler), and FB returns some json

fbchat/utils.py
gave92: I agree it would be simpler. I just didn't like having to remember to put 'true' instead of True xD

Curiously there are some other requests which have a Boolean parameter and those work even without this change.

gave92: BTW the code of the encode_params comes straight from the source of the requests module

madsmtm: Hmm, I think I'll leave it out for now, again because of the added complexity, but if I or somebody else finds another Facebook request where passing `'true'` is *required*, I'll add it back!

madsmtm:
Thanks for the quick response and for the PR!

fbchat/utils.py
madsmtm: Also, I found that there were other issues, eg. when sending images, and a few unicode related bugs in python 2

Issue #260, opened 2018-02-18
Parsing of "Buddylist Overlay"


madsmtm:
In `_parseMessage`, you will recieve a lot of data with the type `buddylist_overlay`. Sample `msg`:
```json
{
  "seq": 50,
  "t": "msg",
  "u": 1234567890,
  "ms": [
    {
      "type": "buddylist_overlay",
      "overlay": {
        "1234567890": {
          "a": 2,
          "c": 4170,
          "s": "push",
          "vc": 74,
          "la": 1518985361
        }
      }
    }
  ]
}
```

The `'a'` (activeness?) value appears in two (and maybe more?) states: `0` and `2`, where `0` means offline and `2` means online.
The `'la'` (last active?) value indicates the timestamp

This can be used to know when users become online or when they go offline (in real time, compared to `onChatTimestamp`, which only recieves these updates once in a while).

If anybody can figure out what the `c`, `s` or `vc` values could mean, please let me know (!), otherwise I'll probably make another `onX` function at some point.

2FWAH:
Hello,

It looks like the NodeJS implementation does not have much more information than we have (https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#listen)

I tried to understand the values of c, s or vc but nothing obvious.
s looks always 'push', c is not always defined, vc is often 74.
Maybe we can get information about those values in Facebook js, but it's not easy.

Do you received those messages without opening your browser?
Even if markAlive is True I need to open my browser to receive buddylist_overlay messages, you too?

madsmtm:
I can't confirm whether or not I can recieve these messages without opening Facebook in my browser, yet. If it's the case, though, then that's really wierd...

**Warning:** There's a lot of complicated code here, if you don't understand boolean operators.

Anyways, I did some digging into the code on `https://www.facebook.com`, it seems like `c` and `s` is unused, and I found that `vc` gets saved, and can later be retrieved per user, with a function they called `c('PresenceStatus').getCapabilities`. It's only used one place, a function they called `c('FBRTCAvailability').isCallable`:
```javascript
function(j) { // j is the user ID here
    var k = c("PresenceStatus").getCapabilities(j); // Gets the `vc` value of the user
    l = c("ChannelConstants").CAPABILITY_VOIP_INTEROP; // this is 8
    return !!(k & l);
}
```

Basically, what I found is that if `vc` is used as some sort of C-style boolean flag-holder-variable:
[...]

PR #273, opened 2018-03-23
Add onBuddylistOverlay method

2FWAH:
Handle Buddylist Overlay messages #260 
At the moment only status and last active timestamp are parsed.

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -1550,6 +1550,13 @@ class Client(object):
                         buddylist[_id] = payload.get('lat')
                     self.onChatTimestamp(buddylist=buddylist, msg=m)
 
+                # Buddylist Overlay
+                elif mtype == "buddylist_overlay":
+                    for user_id in m.get('overlay', {}):
+                        status = UserStatus(m['overlay'][user_id]['a'])
+                        last_active = m['overlay'][user_id]['la']
+                        self.onBuddylistOverlay(user_id=user_id, status=status, last_active=last_active, msg=m)
+
                 # Unknown message type
                 else:
                     self.onUnknownMesssageType(msg=m)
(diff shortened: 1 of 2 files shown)


madsmtm:
Haven't tested it, but I think your implementation looks good. It seems kinda pointless to add this to the API, before we figure out how to make Facebook return the data all the time.
I'm going to leave this open until we do ;)

2FWAH:
I'm still looking why FB does not return overlay messages if the browser is closed...
Don't hesitate to update #260 if you find some clues ;)

2FWAH:
It may be useful to merge the PR to allow a large number of users to test it, thus we could have more feedback to understand how it works.
What do you think about that?

I can add a warning message in the doc: "This is an experimental feature, feedback are welcome".

madsmtm:
I did some more research, seems like Facebook sends multiple requests to `https://[X]-edge-chat.facebook.com/sub` with GET parameters like `sticky_token`, `uid`, `viewer_uid`, `sticky_pool`, and `clientid`

I'm guessing that these requests are tied to the specific sticky_token, and allows different clients to subscribe to different Facebook events.

My theory on the reason we're seeing the different events when `facebook.com` is open in the browser, is that it's probably at Facebook bug or a limitation in their implementation. A kind of side effect, so that when a user is utilising a `sticky_token` (that has requested to see specific events), the events are sent to *all* the user's sessions, regardless of whether the individual `sticky_token`s have subscribed to that event or not.

I don't think adding it to the API will help yet, since without the permanent subscription, it's kinda useless, but I think the steps to proceed are just to analyze the `/sub` request, and make them ourselves in `startListening`. Maybe they need to be made more often than that, too?

madsmtm:
I'll close this since neither of us really investigated the issue any further, and @kapi2289 ended up implementing it in #338 (and the issue with the subscription was fixed in #327). I'm sorry I didn't follow up on this sooner, or at least got your PR merged before #338, I just kinda forgot. Thanks a lot for your work though! 👍

PR #278, opened 2018-04-01
Use datetime objects rather than milliseconds as a string

### LINKED ISSUE #277: Message timestamps seem to be invalid.
Getting the last time stamp of a group (or of a message) gives a string of an abnormally high integer value:

```
>> thread = client.fetchThreadList()[0]
>> ts = int(thread.last_message_timestamp)
>> ts
1522504826129
>> datetime.fromtimestamp(ts)
ValueError: year 50216 is out of range
```

tcmal:
*Fixes #277*

**Changes:**

  - `Message.timestamp` is now a datetime object
  - `Thread.last_message_timestamp` is now a datetime object
  - `before` parameter of `Client.fetchThreadMessages()` & `Client.fetchThreadList()` is now a datetime object
  - *Added `utils.unix_time_millis` and `utils.millis_to_datetime`*

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -10,8 +10,8 @@ from mimetypes import guess_type
 from .utils import *
 from .models import *
 from .graphql import *
-import time
-
+from datetime import datetime
+import calendar
 
 
 class Client(object):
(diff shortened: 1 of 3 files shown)


fbchat/client.py
```diff
@@ -10,8 +10,8 @@
 from .utils import *
 from .models import *
 from .graphql import *
-import time
```
madsmtm: `time` is still being used in `time.sleep`, and neither `datetime` nor `calendar` is being used

PR #285, opened 2018-05-03
v2.0.0 - Api Reformat

madsmtm:
# Tl;dr:
This PR contains several different improvements and changes to the API.
- Caching system, `get` instead of `fetch`
- PEP 8 Style Guide, aka. `on_message` instead of `onMessage`
- [Google Style](http://google.github.io/styleguide/pyguide.html) docstrings, using [napoleon](http://www.sphinx-doc.org/en/stable/ext/napoleon.html)
- `thread` instead of `thread_id` and `thread_type`
- `Client` has been split into several files
These improvements are, however, only an attempt at giving an overview of how I envision the new API, so most of the actual implementation code has been removed for now.
Make sure to take a look in the `examples` folder, to get a feel for how I want the API to look like.

# Motivation
I've been working on reformatting the entire API, and I have quite a lot of small things I want to change. But the big thing, I want it to "just work". That means no more returning an ID of a message, or having `Group` contain a list of user IDs. I want to make a system that's easy to work with, a system where we just return the `Message` object or return a list of `User` objects.

And one of the most tedious things is working with the `fetch_X` methods, where some use the `before` parameter, and some use `offset` parameter. There are very few cases where a `get_X` method with a single `limit` parameter, and where it would just automatically send multiple requests as needed, wouldn't suffice. And imagine, recieving `Thread` objects in `on_X` methods instead of `thread_id` and `thread_type`, `user_id`, `author_id`, and so on.

# Caching
The most important aspect of all of this is making a caching system, where the client would store `Thread` objects, and when needed by a method, just use those. And when updated, eg. when listening, and someone changes a thread's emoji, the `Thread` object could be globally updated, and all places using that `Thread` object would instantly be able to use the new information.

An important factor here is that we need to use the `on_X` calls, so that we can update the `Thread`s, when new information arrives. This could be done (i think) in two ways:
Just override the methods, and then *require* people to make a `super` call before anything else, or, and this is the way I've currently decided on, make private `_on_X` methods, that receive the data first, has a chance to update the cached objects, and then sends the data onwards to `on_X`

# PEP 8 & Split Client
I've also decided to change all method names to lowercase, following the [PEP 8 Style Guide](https://www.python.org/dev/peps/pep-0008/), mostly just to follow *a* style guide. This'll allow us to later use tools like `pycodestyle`, and get a cleaner codebase. Following the "cleaner codebase", I've tried to split the `Client` up into components, so it'll be easier to understand what's going on, and write higher level code.

EDIT:
I have also started using [Google Style](http://google.github.io/styleguide/pyguide.html) docstrings, just because the `sphinx`/`rst` syntax was becoming really terse

I know it's a lot of changes, but if you have the time, I'd really like your input on this. Have a nice day! ;)


examples/advanced_bot.py
```diff
@@ -0,0 +1,69 @@
...
+from fbchat import Client, Group, Message
+
+
+class Bot(Client):
+
+    def on_message(self, message):
+        thread = message.thread
+
+        # Prevent the bot from answering the bot's own messages
+        if message in self.sent_messages:
```
madsmtm: This could be a nice feature to have in general, eg. as `listen(dont_answer_own=True)`, or something like that.
This entire `on_message` could then have been split into several `on_text` and `on_file`

examples/possible_usage.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: UTF-8 -*-
+
+from fbchat import *
+
+c = Client('<email>', '<password>')
+
+# This is a possible way we *could* build the library. Not saying it's a good
+# idea, but I'd like your opinions
```
madsmtm: Leave your feedback about this here

fbchat/__init__.py
```diff
@@ -1,29 +1,53 @@
...
+
+    'Thread',
+    'User',
+    'Group',
+    'Page',
+
+    'Client'
 ]
+
+logging.getLogger(__name__).addHandler(logging.NullHandler())
```
madsmtm: Propper way to add a logger. In files that need to use the logger, `log = logging.getLogger(__name__)` would then be used at the top of the document

fbchat/get.py
```diff
@@ -0,0 +1,154 @@
+# -*- coding: UTF-8 -*-
+
+from __future__ import unicode_literals
+from .base import Base
+
+
+class Get(Base):
+    """Enables retrieving information about threads and messages"""
```
madsmtm: A `Fetch` class would also be made, where most of these methods would have a corresponding `fetch_X` method.

fbchat/get.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def get_threads(self, limit=None):
+        """Retrieve the threads that the client is currently chatting with
+
+        If ``limit`` is ``None``, the result will be a generator, otherwise the
+        result will be a list with maximum length of ``limit``
```
madsmtm: I'm not sure if this is a good idea. Maybe we should just always return a list, or maybe we should remove the ``limit`` parameter and just always return an iterable, where people could shorten it as needed with [`itertools.islice`](https://docs.python.org/library/itertools.html#itertools.islice)

fbchat/get.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+    def get_threads_from_ids(self, *thread_ids):
+        """Retrieve threads based on their IDs
+
+        If ``thread_ids`` contains a single iterable, then a generator yielding
+        the threads is returned.
+
+        If ``thread_ids`` is a single value or multiple values, then a list
+        containing the threads is returned.
```
madsmtm: Not sure if it's a good idea to be so lax about the input to this, and subsequent methods. Maybe we should just only allow a single iterable?

examples/advanced_bot.py
madsmtm: Maybe an `__init__` parameter? `enable/disable` method? Or maybe even a function decorator?
Actually, thinking about it, I think I like the idea of a function decorator for this quite a lot.

fbchat/get.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Args:
+            thread (`Thread`): Thread to fetch files from
+
+        Return:
+            List of `FileAttachment` objects or a generator yielding
+            `FileAttachment` objects
+        """
+
+    def get_url(self, attachment):
```
madsmtm: Maybe `image.get_url(client)` would be better?

madsmtm: Or maybe an internal reference, so you can say `image.get_url()`? Though from the Zen of Python, "Explicit is better than implicit", and you can't easily use two clients in the same script interacting

fbchat/models.py
```diff
@@ -4,471 +4,169 @@
...
+    """
 
-    def __unicode__(self):
-        return '<{} {} ({})>'.format(self.type.name, self.name, self.uid)
+
+class Thread(object):
+    """Represents a Facebook chat-thread
+
+    Attributes:
+        id (int): The unique identifier of the thread
```
madsmtm: `Thread.id` or `Thread.uid`? `Thread.id` doesn't actually conflict with the builtin `id`, and is more clean IMO

madsmtm: Also, I'm advocating for integers, though maybe we should leave it as strings?

fbchat/models.py
```diff
@@ -4,471 +4,169 @@
...
-        self.color = color
-        self.emoji = emoji
+    """Represents a user and the chat-thread the client has with the user
+
+    Attributes:
+        first_name: The user's first name
+        last_name: The user's last name
+        is_friend (bool): Whether the user and the client are friends
+        gender: The user's gender
+        affinity (float): From 0 to 1. How close the client is to the user
```
madsmtm: Maybe a few aliases to `Thread.nicknames`, such as `own_nickname` and `nickname`

fbchat/models.py
```diff
@@ -4,471 +4,169 @@
...
-        self.likes = likes
-        self.sub_title = sub_title
-        self.category = category
+    """Represents a Facebook page
+
+    Attributes:
+        city: The name of the page's location city
+        likes: Amount of likes that the page has
+        sub_title: Some extra information about the page
+        category: The page's category
```
madsmtm: Many of these attributes should probably have a model for themselves

fbchat/models.py
```diff
@@ -4,471 +4,169 @@
...
+
+    Attributes:
+        fb_error_code (int): The error code that Facebook returned
+        fb_error_message: A localized error message that Facebook returned
+    """
 
-    def __unicode__(self):
-        return '<{} {} ({})>'.format(self.type.name, self.name, self.uid)
+
+class Thread(object):
```
madsmtm: This class would override `__eq__`, so that the examples would be valid

fbchat/base.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: UTF-8 -*-
+
+from __future__ import unicode_literals
+from .models import User
+
+
+class Base(User):
```
madsmtm: Is this a good idea? To inherit from the `User` class? It would be what allows us to do the stuff in the examples, such as :
```py
def on_X(self, thread, actor, ...):
    if self != actor:
        ...
    # or
    if self not in thread.admins:
        ...
# or
client.send_text(client, 'Hi me!')
```
Which is a cleaner abstraction from the concept of IDs, and besides, we need an object to signify the client

fbchat/send.py
```diff
@@ -0,0 +1,143 @@
...
+
+from __future__ import unicode_literals
+from .listener import Listener
+from .models import Size
+
+
+class Send(Listener):
+    """Enables sending various messages to threads
+
+    Every method in this class, except `send`, `on_message` and `upload`, are
```
madsmtm: I previously thought this was a bad idea, but after working with it, I realised that it's much more handy to type `client.send_text(thread, 'text')` instead of the more cumbersome `client.send(thread, Message('text'))`.
What are some other aliases you'd like?

fbchat/send.py
```diff
@@ -0,0 +1,143 @@
...
+    shortcuts of the aforementioned
+    """
+
+    #: Contains a list of all messages sent by the client
+    sent_messages = None
+
+    def send(self, thread, message):
```
madsmtm: I'd like to be able to use `self.send(thread, message)` in `on_message`, but in a way where it's fully functional, that means sending images, videos, attachments and so on back. This means that `send` should be able to parse all `Message` objects, and send a proper response.

fbchat/get.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Args:
+            *message_ids: One or more message IDs to query
+
+        Return:
+            `Message` objects, in the order and format their IDs were supplied
+        """
+
+
+    def get_messages(self, thread, limit=None):
```
madsmtm: So, we can retrieve messages, but what about events? By event I mean the small pieces of "history", that eg. says "X left the group". The actions that `ThreadControl` and `ThreadInterraction` can do and see, as they're happening, but we have no way of looking back, and seeing when they happened.

fbchat/thread_interraction.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def set_image(self, thread, image):
+        """Set a group's image
+
+        Args:
+            thread (`Thread`): Group whose image to change
+            image (`Image`): New image
+        """
+
+    def on_image_set(self, thread, actor, old_image):
```
madsmtm: The `old_x` can be quite useful, but if we haven't cached the thread before, we won't have a way of knowing this? How should this state be signified? Should we just pass ``None``, and then document that this instance might happen?

madsmtm: Also notice how I've changed it from the way it was before, so that the passed `thread` object automatically contains the updated information

fbchat/models.py
```diff
@@ -4,471 +4,169 @@
 import enum
```
madsmtm: Currently, the models are still collected in a single file, but perhaps it would make more sense to spread them out into the other files, like `Message` would be in `send.py`?

PR #286, opened 2018-05-07
Fix `IndexError: list index out of range` bug

orenyomtov:
When the returned `short_name` is null, `fbchat` throws an exception:

```python
  File "/usr/local/lib/python2.7/site-packages/fbchat/client.py", line 792, in fetchThreadList
    return [graphql_to_thread(node) for node in j['viewer']['message_threads']['nodes']]
  File "/usr/local/lib/python2.7/site-packages/fbchat/graphql.py", line 193, in graphql_to_thread
    last_name=user.get('name').split(user.get('short_name'),1)[1].strip(),
IndexError: list index out of range
```

This commit fixes that scenario by accessing the last item in the list via `.pop()` instead of via `[1]`

### DIFF
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -190,7 +190,7 @@ def graphql_to_thread(thread):
             url=user.get('url'),
             name=user.get('name'),
             first_name=user.get('short_name'),
-            last_name=user.get('name').split(user.get('short_name'),1)[1].strip(),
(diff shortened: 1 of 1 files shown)


madsmtm:
I'm pretty sure it works when `short_name` is null/`None`, but when it's a string that's not in `user['name']`, then it fails. You're right though, `pop` is the best option in all cases

PR #287, opened 2018-05-08
Make obtaining fb_h and fb_dtsg more versatile

orenyomtov:
fb_dtsg is sometimes returned inside an HTML comment, and beautifulsoup can't find it - in that case we'll use a regular expression to extract it.

fb_h is sometimes not returned in the HTML of req_url.BASE (in my experience, when resuming a session using session_cookies).

Following the discussion here:
Schmavery/facebook-chat-api#505
I learned it is used for logging out, and can be found in the response of `https://www.facebook.com/bluebar/modern_settings_menu/`.

I included support for fetching it from there.

Because this library is used many more times for logging in, than for logging out, instead of adding an extra HTTP request during login, I decided to perform it during logout, only in case fb_h was not previously found in the HTML of req_url.BASE.

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -209,8 +209,18 @@ class Client(object):
 
         r = self._get(self.req_url.BASE)
         soup = bs(r.text, "lxml")
-        self.fb_dtsg = soup.find("input", {'name':'fb_dtsg'})['value']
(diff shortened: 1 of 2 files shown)


madsmtm:
Good call on not force-fetching the `fb_h` before it's actually needed.

PR #290, opened 2018-05-16
Fix: Resolve FBChatException

Abhinav2812:
Resolve the error `fbchat.models.FBchatException: Could not get ThreadColor from color: FF0084FF` when threadcolor is set to default (MESSENGER_BLUE)

### DIFF
--- a/fbchat/models.py
+++ b/fbchat/models.py
@@ -469,7 +469,7 @@ class EmojiSize(Enum):
 
 class ThreadColor(Enum):
     """Used to specify a thread colors"""
-    MESSENGER_BLUE = ''
+    MESSENGER_BLUE = '#0084ff'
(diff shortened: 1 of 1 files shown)


madsmtm:
Thanks for the submission, Facebook must have changed the value at some point

Abhinav2812:
Glad I could be of help :smile:

madsmtm:
Though changing this fixed the issues with `fetchUserInfo`, `fetchThreadList` and also `onColorChange`, it broke `changeThreadColor`. It's fixed since a4dfe0d279569f76b86566195b30a8b7beb3d851 ;)

The thing that made me miss this is that Facebook doesn't complain if the color passed via. `changeThreadColor` is invalid, it just doesn't update the color either

Issue #220, opened 2017-10-19
ThreadColor Enums Error


ekohilas:
Currently, The ThreadColor Enums only support a given set of colours, however previous versions of messenger allowed for any thread colour to be set through a post request, and so currently the code breaks when thread info is called for such a thread.

There's a couple of ways this could be solved:
* Whenever the thread color is passed back to the user, it should be passed back as a string instead. The Enum would still exist to allow the user to send the correct data when changing the thread colour. This feels like the more correct option and would be better for the future, but would break backwards compatibility (Which wouldn't be too bad as other improvements could be made too)
* Make the type of all thread colours strings
* Whenever a new ThreadColor is seen, the enum could be extended to include it
https://stackoverflow.com/a/35899963. This wouldn't break backwards compatibility, but would require the aenum package, and seems more hacky.

madsmtm:
I see that that's a problem, but I'd really like to avoid breaking backwards compatability for as long as possible. Maybe we should just log the exception instead of raising it, and then have the threadcolor be `None` or a string? This would be a tempoary fix, until we decide to break backwards compatability anyways.
When we do decide to do so, please remind me, and make this system better ;)

PR #293, opened 2018-05-26
Added missing color

feelfreelinux:
This fixes `models.FBchatException: Could not get ThreadColor from color: FF00C517` exception that I ran into while trying to use this library.


madsmtm:
This is the same issue as described in #220. And we haven't made a fix for it yet, though it's probably not that complicated

PR #295, opened 2018-06-01
Fix onTyping detection

2FWAH:
FB changed the format of typing notification messages:
- update "mtype" from "typ" to "ttyp".
- Get thread ID from "to" to "thread_fbid" ("thread" looks the same)

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -1529,9 +1529,9 @@ class Client(object):
                     self.onInbox(unseen=m["unseen"], unread=m["unread"], recent_unread=m["recent_unread"], msg=m)
 
                 # Typing
-                elif mtype == "typ":
+                elif mtype == "ttyp":
                     author_id = str(m.get("from"))
-                    thread_id = str(m.get("to"))
+                    thread_id = str(m.get("thread_fbid"))
                     if thread_id == self.uid:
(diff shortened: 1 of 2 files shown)


madsmtm:
I looked at it a bit, and you're partially right, though I found that `mtype` is set to `typ` when its a user thread, and `ttyp` when it's a group thread.

Apart from that, the `to` parameter were incorrectly used as `thread_id` before for user chats anyway, because when it's a user chat you would expect `thread_id` to be the other person's ID, but `to` would be your own ID.

2FWAH:
I fixed my PR to call onTyping for both "typ" and "ttyp".

madsmtm:
I realised that I couldn't properly explain the issue, so I just went ahead and fixed it. `thread_fbid` is only available to group threads, but on user threads, `to` is still wrong to use, since it simply points to yourself. There (unless you're the person sending the typing status) `author_id` would be the correct user thread, too.

madsmtm:
I realised that I couldn't properly explain the issue, so I just went ahead and fixed it. `thread_fbid` is only available to group threads, but on user threads, `to` is still wrong to use, since it simply points to yourself. There (unless you're the person sending the typing status) `author_id` would be the correct user thread, too.

PR #296, opened 2018-06-03
TravisCI integration and updated test suite

madsmtm:
I've changed the test suite from using `unittest` to using `pytest`, and reworked it, so that we can test when `onX` events were successfully executed. Though I haven't reworked all parts, just adding these extra tests lead me to the discovery of several bugs and unexpected behaviour.

There's admittedly great part of the test code that still has redundancy, but it's getting closer to an acceptable state.

I've tinkered with TravisCI configuration, and I think it's now at a point where it's pretty stable. The main reason for using TravisCI is primarily to detect python 2/3 issues, and changes on Facebook's side.

So, @carpedm20, I need you to setup a Travis project and add either a Github integration or a webhook, and then set the following settings:
```
Build only if .travis.yml is present -> true
Limit concurrent jobs -> true, set to 1 (This one is the most important, since the clients can't be used by two jobs at once)
Build pushed branches -> true
Build pushed pull requests -> true
Auto Cancellation (Both) -> true
Cron job: master, daily, "Do not run if there has been a build in the last 24h"
```

I'd also like you to setup a [RTFD Webhook](https://docs.readthedocs.io/en/latest/webhooks.html) with Github, so the docs are automatically built, just something I never got around to asking you to do. You should have access to the project on [RTFD](https://readthedocs.org/projects/fbchat/)

Lastly you can, if you want to, set some branch protections on master, preventing force pushes definitely should be enabled, but you can choose to require status checks and such.

Note: I initally set this off to v2.0.0, but I decided against that, and made it now, so that the jump to v2.0.0 is smaller


carpedm20:
@madsmtm I think I set up a Travis project. Let me know if it doesn't work

madsmtm:
Yup, it works flawlessly, thanks ;)

madsmtm:
Hey @carpedm20, I looked at it again, and from what I understand it seems like you're using the paid version of TravisCI, and to that I have to say something; if you don't use it for anything else, you should use the free version, since this is an open source project, and the free version was made specifically for that ;)

carpedm20:
@madsmtm Thanks! I think I removed paid TravisCI integration for my account. Now I understand the difference between `travis-ci.com` and `travis-ci.org`. Let me know if anything works in a wrong way :)

madsmtm:
Hi @carpedm20, I might have `@mentioned` you incorrectly, so I'll do it again

carpedm20:
Sorry, @madsmtm. I think Travis-ci works well as:
image

If you mean by webhook for RTFD, I just added it. Let me know again if things don't work yet.

madsmtm:
The "CRON" service works fine, and I can go in and use Travis' "Trigger Build" functionality, but the problem is not that the builds are failing, it's that they're not automatically activating, when I push. Aka. Travis isn't receiving updates from Github automatically

madsmtm:
@carpedm20, the "CRON" service works fine, and I can go in and use Travis' "Trigger Build" functionality, but the problem is not that the builds are failing, it's that they're not automatically activating, when I push. Aka. Travis isn't receiving updates from Github automatically.
But no need to apologise about anything ;)

carpedm20:
@madsmtm I see. I wonder whether we should add `notifications: _webhooks: _` to `.travis.yml` as noted in https://docs.travis-ci.com/user/notifications/#Configuration-in-Travis-CI

madsmtm:
Okay, @carpedm20, I found this: https://docs.travis-ci.com/user/open-source-on-travis-ci-com/, which I didn't know about before.
As far as I understand, there's two components:
- [Travis CI Github Service](https://github.com/github/github-services/blob/master/lib/services/travis.rb), which is being deprecated, but is the way to access repos on travis-ci.org
- [Travis CI Github App](https://github.com/marketplace/travis-ci), which is the way to access repos on travis-ci.com

Here's a picture showing the two:


But as noted above, the Github Service is being deprecated, and so is travis-ci.org.
So, I think it was correct to have our project on travis-ci.com, of course with the free, open source plan. That means you should setup the Github App, and disable the Github Service, and perhaps look at https://travis-ci.org/profile and https://travis-ci.com/profile, to make sure you disable the project at travis-ci.org, and enable it on travis-ci.com again.
Here's how I _think_ it should look when you're done:


Don't worry about the settings on [the project](https://travis-ci.com/carpedm20/fbchat/settings), I can apparently change those, since I'm a contributor

carpedm20:
I see. That means, the previous setting of `travis-ci.com` was correct, I guess? I disabled GitHub services and activated `travis-ci.com`. The problem is, because `travis-ci.org` is activated, it can't be migrated to `travis-ci.com` now though there is no service?

image

image

madsmtm:
That's annoying... Well, @carpedm20, is there any way we can delete the `travis-ci.org` version? Or should we contact Travis Support?
> Repositories may also be migrated without their build history or build settings (including environment variables) immediately. Please contact support support@travis-ci.com and include “Repository Migration” somewhere in the subject line, and the name of the repository in your email.

Another option is to try enabling the Github Service, and the project on `travis-ci.org`, and then just wait until Travis migrates it?
> However, open source repositories will be migrated to travis-ci.com gradually, beginning at the end of Q2 2018. You will receive an email when the migration for a repository is complete. This is an opt-in process: to have a repository migrated over, it must first be activated on travis-ci.com.

carpedm20:
Okay, I enabled Github Service back now.

madsmtm:
I've just had the time to test it now, and it works great, thanks!

Issue #294, opened 2018-05-26
Method to remove friends?


DeltaF1:
I can't find anything in the documentation about removing friends. Does anyone know what the ajax url is, or is this something that is yet to be investigated?

madsmtm:
Removing and adding friends, and receiving frend requests is not yet part of the API, feel free to investigate ;)

ThatAlexanderA:
I searched a little bit and I found that https://m.facebook.com/removefriend.php?friend_id=user_id this url is to remove friend. But you need to click the confirm button to remove the friend. The button name is submit. I don't know how to send a requests that simulates the button press.

ThatAlexanderA:
I added the new url to the utils.py
a264fac
Then added your code to the client.py
c842be3
And run the code, it run successfully with no errors. But the friend isn't removed. 

With Firefox dev tools and track down what Post request need to send. And If I'm right we need to send
headers too with the following information:
Content-Type: application/x-www-form-urlencoded
Cookie: c_user=c_user; xs=xs
With the self._session.cookies.get_dict().get('') command we can get them.
`                {
                "name": "Content-Type",
                "value": "application/x-www-form-urlencoded"
                },
                {
                "name": "Cookie",
                "value": ""
                },
                {
                "name": "c_user",
                "value": self._session.cookies.get_dict().get('c_user')
                },
                {
                "name": "xs",
                "value": self._session.cookies.get_dict().get('xs')
                }`
Should I add this code to the request too? And If so where?

madsmtm:
I think `fb_dtsg` is already available from login via. the `Client.fb_dtsg` value, and when using `Client._post`, the cookies are automatically applied to the request, along with a proper `Content-Type`. Maybe there's something else at play?

DeltaF1:
Is it possible that both the initial GET request you found _and_ the
confirmation request I found need to be sent? Alternatively, maybe the
request I found is wrong.

ThatAlexanderA:
I think your url is good. This is the request that is sent: (there are more lines, but if I remove them it still works)
(https://imgur.com/PThXsNZ)
I try without my url and use `Client.fb_dtsg` as the fb_dtsg value, maybe that will work.

madsmtm:
[...]

PR #298, opened 2018-06-04
Added remove friend

ThatAlexanderA:
- Added remove friend url to the utils.py
- Added removeFriend def to the client.py

#294


fbchat/client.py
```diff
@@ -467,8 +467,16 @@ def resetDefaultThread(self):
 
     """
     END DEFAULT THREAD METHODS
-    """
-
+    """ 
+    def removeFriend(self, friend_id=None):
+        """param friend_id: The id of the friend that you want to remove"""
```
madsmtm: Perhaps you could add a short description of the method, along with a documentation of the return value?

fbchat/client.py
```diff
@@ -467,8 +467,16 @@ def resetDefaultThread(self):
...
+        payload = {
+            "friend_id": friend_id,
+            "unref": "none",
+            "confirm": "Confirm",
+        }
+        r = self._post(self.req_url.REMOVE_FRIEND,payload)
+        return r.ok
```
madsmtm: If `r.ok` is not true, an exception could have been thrown by using `fix_request=False` in `Client._post`. Maybe the return value should be used for detecting whether the removal of the person _actually_ succeeded? I tested four scenarios:
If the client isn't logged in, it's redirected to the login page
If the client is logged in, but hasn't been active in a while, it's presented with an error message (No redirect)
If the operation was successful, the client is redirected to `/removefriend.php?friend_id=[friend_id]&removed&__req=e&_rdr`
And if the friend_id was invalid, it's redirected to `/removefriend.php?friend_id=[friend_id]&removed&err=1431010&__req=e&_rdr`

So, I think we should update `_post` to support the parameter `allow_redirects`, just like `_cleanGet` does, then use that, look at `r['Location']`, and determine whether the redirect url starts with `/removefriend`. If it does, but contains the `err` parameter, we can either throw an error, or just return `False`.

fbchat/client.py
ThatAlexanderA: Yes, I will add it soon.

fbchat/client.py
ThatAlexanderA: I allowed redirects in the _post, but I can't get the location, only with `r.url`. Can I use that?

madsmtm: The point of disabling redirects and using `r['Location']` would be to reduce the number of requests that we need to send, but you can start by implementing it without, and just use `r.url`.

Considering testing the error code, though your described method would probably work, I would argue we should use two functions from [`urllib.parse`](https://docs.python.org/3/library/urllib.parse.html), `parse_qs` and `urlparse`, and look at whether the `dict` `parse_qs(urlparse(url).query)` contains an error code.

ThatAlexanderA:
I updated the code. I used print for tell if the remove was successful, but we can use something different if you want.

fbchat/client.py
```diff
@@ -469,14 +474,25 @@ def resetDefaultThread(self):
...
+        :returns True when the removing was successful
+        """
         payload = {
             "friend_id": friend_id,
             "unref": "none",
             "confirm": "Confirm",
         }
         r = self._post(self.req_url.REMOVE_FRIEND,payload)
-        return r.ok
+        dict = parse_qs(urlparse(r.url).query)
```
madsmtm: Be careful about overwriting `dict`, since its a builtin. Use a different, perhaps even a more descriptive name, like `query`

fbchat/client.py
```diff
@@ -469,14 +474,25 @@ def resetDefaultThread(self):
...
+        if "err" not in dict:
+            print ("Remove was successful!")
```
madsmtm: If you feel like it's needed, log output can be useful, but don't use `print`, since people can't selectively disable the log output. Use `log.info` or `log.debug` here instead

fbchat/client.py
```diff
@@ -469,14 +474,25 @@ def resetDefaultThread(self):
     END DEFAULT THREAD METHODS
     """ 
     def removeFriend(self, friend_id=None):
-        """param friend_id: The id of the friend that you want to remove"""
+        """
+        Removes a specifed friend from your friend list
+        :param friend_id: The id of the friend that you want to remove
+        :returns error when the removing was unsuccessful
+        :returns True when the removing was successful
```
madsmtm: The syntax in the docstrings is quite important, since it gets executed and published on [RTFD](https://fbchat.readthedocs.io). `:returns` is wrong, it should be `:return:`, and there should only be one of them

fbchat/client.py
```diff
@@ -11,7 +11,12 @@
 from .models import *
 from .graphql import *
 import time
-
+try:
+    from urllib.parse import urlparse
+    from urllib.parse import parse_qs
+except ImportError:
+    from urlparse import urlparse
+    from urlparse import parse_qs
```
madsmtm: When importing two things from the same module you can use commas, like this: `from urlparse import urlparse, parse_qs`

fbchat/client.py
madsmtm: Just noticed that you addressed this in your comment

ThatAlexanderA:
I updated the code. :)

madsmtm:
Looks good, I'll merge it as soon as I get TravisCI working again

ThatAlexanderA:
Okay, thank you! :)

Issue #250, opened 2018-01-27
Create Group


Marekkon5:
How can I create group? Thanks

madsmtm:
This isn't possible yet, I'll add it to the list ;)

madsmtm:
Fixed in #316, in release `v1.4.0`. The method is called `createGroup`

PR #316, opened 2018-07-14
Added Create Group

ThatAlexanderA:
Added create group url to the utils.py
Added create group def to the client.py
#250


madsmtm:
Thanks for your work, and sorry I've been late on answering, I've been on vacation with very limited internet access.

PR #317, opened 2018-07-15
Fix 2FA for non english FB

gave92:
This commit fixes the 2FA. Currently we check for the string "enter security code to continue" which is only there for the english page of FB.


madsmtm:
Thanks, I'll take your word for it that it works, and sorry I've been late on answering, I've been on vacation with very limited internet access.

PR #326, opened 2018-08-30
Merge `Room` with `Group`

kapi2289:
### Description:
Because [Facebook](https://facebook.com/) connected rooms with groups, I'm doing the same in `fbchat`!
I removed `Room` model and added new parameters to `Group` model:
> admins
> approval_mode
> approval_requests
> join_link

### DIFF
--- a/fbchat/models.py
+++ b/fbchat/models.py
@@ -102,8 +102,16 @@ class Group(Thread):
     color = None
     #: The groups's default emoji
     emoji = None
+    # Set containing user IDs of thread admins
+    admins = None
+    # True if users need approval to join
+    approval_mode = None
+    # Set containing user IDs requesting to join
+    approval_requests = None
+    # Link for joining group
+    join_link = None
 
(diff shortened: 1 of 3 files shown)


fbchat/models.py
```diff
@@ -531,7 +519,6 @@ class ThreadType(Enum):
     USER = 1
     GROUP = 2
     PAGE = 3
```
madsmtm: We should keep `ThreadType.ROOM`, just change it's value to `2`, like `ThreadType.GROUP`, for backwards compatibility

fbchat/models.py
```diff
@@ -114,24 +122,6 @@ def __init__(self, uid, participants=None, nicknames=None, color=None, emoji=Non
         self.nicknames = nicknames
         self.color = color
         self.emoji = emoji
-
-
-class Room(Group):
```
madsmtm: For backwards compatibility, we should add `Room = Group`

PR #333, opened 2018-09-12
[Feature] Extensible attachments

kapi2289:
## Description
Added:
- Detecting extensible attachments
- New models: `ShareAttachment`, `LocationAttachment`, `LiveLocationAttachment`
- Fetching live user location
- Unreacting to messages
- New methods `sendLocation` and `sendPinnedLocation`
- New `on` methods: `onReactionAdded`, `onReactionRemoved`, `onBlock`, `onUnblock`, `onLiveLocation`
- Fixed `size` of attachments
- Sending voice clips
- Unsending messages
- `deleted` Message attribute

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -1150,7 +1150,7 @@ class Client(object):
         :raises: FBchatException if request failed
         """
         self._sendLocation(location=location, current=True, thread_id=thread_id, thread_type=thread_type)
-        
+
     def sendPinnedLocation(self, location, thread_id=None, thread_type=None):
         """
         Sends a given location to a thread as a pinned location
@@ -1572,7 +1572,15 @@ class Client(object):
 
         j = self._post(self.req_url.THREAD_EMOJI, data, fix_request=True, as_json=True)
 
-    def _react(self, message_id, reaction=None):
+    def reactToMessage(self, message_id, reaction):
+        """
+        Reacts to a message, or removes reaction
+
+        :param message_id: :ref:`Message ID <intro_message_ids>` to react to
+        :param reaction: Reaction emoji to use, if None removes reaction
+        :type reaction: models.MessageReaction or None
+        :raises: FBchatException if request failed
+        """
         data = {
             "doc_id": 1491398900900362,
             "variables": json.dumps({
@@ -1585,35 +1593,7 @@ class Client(object):
                 }
             })
         }
-
-        r = self._post(self.req_url.MESSAGE_REACTION, data)
-        r.raise_for_status()
-
-    def reactToMessage(self, message_id, reaction):
-        """
-        Deprecated. Use :func:`fbchat.Client.addReaction` instead
-        """
-        self.addReaction(message_id=message_id, reaction=reaction)
-
-    def addReaction(self, message_id, reaction):
-        """
-        Reacts to a message
-
-        :param message_id: :ref:`Message ID <intro_message_ids>` to react to
-        :param reaction: Reaction emoji to use
-        :type reaction: models.MessageReaction
-        :raises: FBchatException if request failed
-        """
-        self._react(message_id=message_id, reaction=reaction)
-
-    def removeReaction(self, message_id):
-        """
-        Removes reaction from a message
-
-        :param message_id: :ref:`Message ID <intro_message_ids>` to remove reaction from
-        :raises: FBchatException if request failed
-        """
-        self._react(message_id=message_id)
+        self._post(self.req_url.MESSAGE_REACTION, data, fix_request=True, as_json=True)
 
     def createPlan(self, plan, thread_id=None):
         """
@@ -2391,7 +2371,7 @@ class Client(object):
                                     location = graphql_to_live_location(l)
                                     self.onLiveLocation(mid=mid, location=location, author_id=author_id, thread_id=thread_id,
                                                         thread_type=thread_type, ts=ts, msg=m)
-                            
+
                             # Message deletion
                             elif d.get('deltaRecallMessageData'):
                                 i = d['deltaRecallMessageData']
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -1116,6 +1116,26 @@ class Client(object):
             data['specific_to_list[0]'] = "fbid:{}".format(thread_id)
         return self._doSendRequest(data)
 
+    def sendLocation(self, location, thread_id=None, thread_type=None):
+        """
+        Sends a given location to a thread
+
+        :param location: Location to send
+        :param thread_id: User/Group ID to send to. See :ref:`intro_threads`
+        :param thread_type: See :ref:`intro_threads`
+        :type location: models.LocationAttachment
+        :type thread_type: models.ThreadType
+        :return: :ref:`Message ID <intro_message_ids>` of the sent message
+        :raises: FBchatException if request failed
+        """
+        thread_id, thread_type = self._getThread(thread_id, thread_type)
+        data = self._getSendData(thread_id=thread_id, thread_type=thread_type)
+        data['action_type'] = 'ma-type:user-generated-message'
+        data['location_attachment[coordinates][latitude]'] = location.latitude
+        data['location_attachment[coordinates][longitude]'] = location.longitude
+        data['location_attachment[is_current_location]'] = True
+        return self._doSendRequest(data)
+
     def _upload(self, files):
         """
         Uploads files to Facebook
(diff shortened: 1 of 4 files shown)


fbchat/utils.py
```diff
@@ -297,3 +298,10 @@ def get_files_from_paths(filenames):
     yield files
     for fn, fp, ft in files:
         fp.close()
+
+def get_url_parameters(url, *args):
+    params = parse_qs(urlparse(url).query)
+    return [params[arg][0] for arg in args if params.get(arg)]
+
+def get_url_parameter(url, param):
+    return get_url_parameters(url, param)[0]
```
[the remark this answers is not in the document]
kapi2289: I think we should keep this methods for future, and btw I think it's a lot of cleaner to make separate methods. :blush:

fbchat/client.py
```diff
@@ -2792,6 +2812,79 @@ def onGamePlayed(self, mid=None, author_id=None, game_id=None, game_name=None, s
         """
         log.info("{} played \"{}\" in {} ({})".format(author_id, game_name, thread_id, thread_type.name))
 
+    def onReactionAdded(self, mid=None, reaction=None, add_reaction=None, author_id=None, thread_id=None, thread_type=None, ts=None, msg=None):
```
madsmtm: Does `add_reaction` have any meaning when we have `onReactionRemoved`?

fbchat/client.py
```diff
@@ -2792,6 +2812,79 @@ def onGamePlayed(self, mid=None, author_id=None, game_id=None, game_name=None, s
...
+        :param author_id: The ID of the person who removed reaction
+        :param thread_id: Thread ID that the action was sent to. See :ref:`intro_threads`
+        :param thread_type: Type of thread that the action was sent to. See :ref:`intro_threads`
+        :param ts: A timestamp of the action
+        :param msg: A full set of the data recieved
+        :type thread_type: models.ThreadType
+        """
+        log.info("{} removed reaction from {} message in {} ({})".format(author_id, mid, thread_id, thread_type))
+
+    def onBlock(self, author_id=None, thread_id=None, thread_type=None, ts=None, msg=None):
```
madsmtm: Does `thread_id` and `thread_type` have any meaning here? If so, the documentation should explain that. Same goes for `onUnblock`

kapi2289: This method executes too when you block somebody, then `author_id` is yours, and `thread_id` is the blocked user ID.

fbchat/graphql.py
```diff
@@ -128,9 +128,76 @@ def graphql_to_attachment(a):
...
+    if story:
+        _type = story['target']['__typename']
+        if _type == 'MessageLocation':
+            latitude, longitude = get_url_parameter(get_url_parameter(story['url'], 'u'), 'where1').split(", ")
+            rtn = LocationAttachment(
+                uid=int(story['deduplication_key']),
+                latitude=float(latitude),
+                longitude=float(longitude),
+            )
+            rtn.image_url = story['media']['image']['uri']
```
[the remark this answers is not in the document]
kapi2289: `LocationAttachment.__init__()` is created for users, so they can use this model for `sendLocation`

fbchat/client.py
```diff
@@ -1116,6 +1116,26 @@ def wave(self, wave_first=True, thread_id=None, thread_type=None):
...
+        :type thread_type: models.ThreadType
+        :return: :ref:`Message ID <intro_message_ids>` of the sent message
+        :raises: FBchatException if request failed
+        """
+        thread_id, thread_type = self._getThread(thread_id, thread_type)
+        data = self._getSendData(thread_id=thread_id, thread_type=thread_type)
+        data['action_type'] = 'ma-type:user-generated-message'
+        data['location_attachment[coordinates][latitude]'] = location.latitude
+        data['location_attachment[coordinates][longitude]'] = location.longitude
+        data['location_attachment[is_current_location]'] = True
```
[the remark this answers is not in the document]
kapi2289: When True, title is `User's location`, when False it's `Pinned location`. We can make `pinned` attribute or make separate method.

fbchat/client.py
madsmtm: But since `thread_type` always will be `ThreadType.USER`, and `thread_id` is duplicated, is there any reason for having this?

fbchat/client.py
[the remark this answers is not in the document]
madsmtm: I think I'll vote for making a seperate method

fbchat/client.py
kapi2289: `thread_type` we can remove, but `thread_id` is not always duplicated as I wrote.

madsmtm: Oh, sorry, I misread, you're right ;)

But maybe `subject_id` is more descriptive, then?

fbchat/client.py
```diff
@@ -1485,29 +1572,48 @@ def changeThreadEmoji(self, emoji, thread_id=None):
...
                     "message_id": str(message_id),
-                    "reaction": reaction.value
+                    "reaction": reaction.value if reaction else None
                 }
             })
         }
 
-        j = self._post(self.req_url.MESSAGE_REACTION, full_data, fix_request=True, as_json=True)
+        r = self._post(self.req_url.MESSAGE_REACTION, data)
+        r.raise_for_status()
```
madsmtm: Use `fix_request` in `Client._post` instead of `r.raise_for_status()`. And when the response is actually json, use `as_json` as well, that allows for better error handling

fbchat/client.py
```diff
@@ -1485,29 +1572,48 @@ def changeThreadEmoji(self, emoji, thread_id=None):
...
+
+    def reactToMessage(self, message_id, reaction):
+        """
+        Deprecated. Use :func:`fbchat.Client.addReaction` instead
```
madsmtm: Is it really nescessary to deprecate this? I think using `client.reactToMessage("<message_id>", None)` is pretty expressive

fbchat/graphql.py
```diff
@@ -131,47 +131,55 @@ def graphql_to_attachment(a):
...
-            )
-            rtn.image_url = story['media']['image']['uri']
-            rtn.image_width = story['media']['image']['width']
-            rtn.image_height = story['media']['image']['height']
-            rtn.url = story['url']
-            return rtn
-        elif _type in ['ExternalUrl', 'Story']:
-            return ShareAttachment(
+        target = story.get('target')
+        if target:
```
madsmtm: Is there nothing else that can determine whether the attachment is a `DeletedMessage`?

fbchat/models.py
```diff
@@ -216,6 +216,12 @@ def __init__(self, uid=None):
         """Represents a Facebook attachment"""
         self.uid = uid
 
+class DeletedMessage(Attachment):
```
madsmtm: How is the `DeletedMessage` event an attachment? Is it attached to the message which is deleted? If so, why not just make a `deleted` attribute on the message?

fbchat/client.py
kapi2289: Well, so why `sendMessage` is marked as deprecated?

fbchat/graphql.py
kapi2289: No, just all information about the attachment is `null`

fbchat/models.py
kapi2289: It's weird, but it is an attachment, but I'll try somehow to make `deleted` attribute

fbchat/client.py
madsmtm: By this I also meant that we don't really need `_react`, `addReaction` and `removeReaction`

fbchat/client.py
```diff
@@ -2400,26 +2380,29 @@ def getThreadIdAndThreadType(msg_metadata):
...
+                        deleted = False
                         if delta.get('attachments'):
                             try:
                                 for a in delta['attachments']:
                                     mercury = a['mercury']
                                     if mercury.get('blob_attachment'):
                                         image_metadata = a.get('imageMetadata', {})
                                         attach_type = mercury['blob_attachment']['__typename']
-                                        attachment = graphql_to_attachment(mercury.get('blob_attachment', {}))
+                                        attachment = graphql_to_attachment(mercury['blob_attachment'])
```
madsmtm: Are you sure this change is safe? I don't remember if there was a good reason for making it like it is 😕

kapi2289: Of course!
- `[]` are used one line up anyways.
- There's an `if` checking if a `'blob_attachment'` key exists, so there's no reason to 'verify' it again.
- If `.get()` would return `{}`, `graphql_to_attachment` will raise the error anyways.

madsmtm: Right, I didn't even spot the `if` statement above... :grin:

madsmtm:
🎉

PR #338, opened 2018-09-22
[Feature] Active status

kapi2289:
## Description
- Added `ActiveStatus` model
- Added changing client active status while listening
- Added fetching friends' active status
- Fixed `onImageChange` documentation and added missing `msg` parameter
- Moved `on` methods to the right place

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -62,6 +62,8 @@ class Client(object):
         self.default_thread_id = None
         self.default_thread_type = None
         self.req_url = ReqUrl()
+        self._markAlive = True
+        self._buddylist = dict()
 
         if not user_agent:
             user_agent = choice(USER_AGENTS)
@@ -989,6 +991,15 @@ class Client(object):
         plan = graphql_to_plan(j["payload"])
         return plan
 
+    def getUserActiveStatus(self, user_id):
+        """
+        Gets friend active status as an :class:`models.ActiveStatus` object
+
+        :param user_id: ID of the user
+        :rtype: models.ActiveStatus
+        """
+        return self._buddylist.get(user_id)
+
     """
     END FETCH METHODS
     """
@@ -1747,7 +1758,7 @@ class Client(object):
         .. todo::
             Documenting this
         """
-        r = self._post(self.req_url.MARK_SEEN, {"seen_timestamp": 0})
+        r = self._post(self.req_url.MARK_SEEN, {"seen_timestamp": now()})
         return r.ok
 
     def friendConnect(self, friend_id):
@@ -2090,7 +2101,7 @@ class Client(object):
                                 image_metadata = fetch_data.get("image_with_metadata")
                                 image_id = int(image_metadata["legacy_attachment_id"]) if image_metadata else None
                                 self.onImageChange(mid=mid, author_id=author_id, new_image=image_id, thread_id=thread_id,
-                                                   thread_type=ThreadType.GROUP, ts=ts)
+                                                   thread_type=ThreadType.GROUP, ts=ts, msg=m)
 
                     # Nickname change
                     elif delta_type == "change_thread_nickname":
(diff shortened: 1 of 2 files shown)


fbchat/client.py
```diff
@@ -1747,7 +1758,7 @@ def markAsSeen(self):
         .. todo::
             Documenting this
         """
-        r = self._post(self.req_url.MARK_SEEN, {"seen_timestamp": 0})
+        r = self._post(self.req_url.MARK_SEEN, {"seen_timestamp": now()})
```
madsmtm: Did you get `markAsSeen` to work?

fbchat/client.py
```diff
@@ -989,6 +991,15 @@ def fetchPlanInfo(self, plan_id):
         plan = graphql_to_plan(j["payload"])
         return plan
 
+    def getUserActiveStatus(self, user_id):
```
madsmtm: We need to include some information about that we can only get this information when the listener loop is running

fbchat/client.py
```diff
@@ -2397,21 +2425,33 @@ def stopListening(self):
...
-    def listen(self, markAlive=True):
+    def listen(self, markAlive=None):
         """
         Initializes and runs the listening loop continually
 
         :param markAlive: Whether this should ping the Facebook server each time the loop runs
         :type markAlive: bool
         """
+        if markAlive is not None:
+            self._markAlive = markAlive
```
madsmtm: This should use `Client.setActiveStatus`

fbchat/client.py
kapi2289: I don't really know what mark as seen is supposed to do :joy: but I saw in network tools that the sent timestamp is just actual timestamp, so I just changed it.

kapi2289:
Yes, I saw that, this was fixed by #327 :wink:

fbchat/client.py
madsmtm: Haha, okay, fair enough 👍

fbchat/models.py
```diff
@@ -520,6 +520,25 @@ def __repr__(self):
     def __unicode__(self):
         return '<Plan ({}): {} time={}, location={}, location_id={}>'.format(self.uid, repr(self.title), self.time, repr(self.location), repr(self.location_id))
 
+class ActiveStatus(object):
+    #: Whether the user is active now
+    active = None
+    #: Timestamp when the user was last active
+    last_active = None
```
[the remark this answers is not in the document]
madsmtm: Or maybe just keep it as is for consistency with the other values around the code

fbchat/client.py
```diff
@@ -2335,12 +2349,27 @@ def getThreadIdAndThreadType(msg_metadata):
 
                 # Chat timestamp
                 elif mtype == "chatproxy-presence":
-                    buddylist = {}
+                    buddylist = dict()
                     for _id in m.get('buddyList', {}):
                         payload = m['buddyList'][_id]
-                        buddylist[_id] = payload.get('lat')
+                        last_active = payload.get('lat')
+                        active = time.time() - last_active <= 60
```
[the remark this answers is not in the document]
kapi2289: It was just guessing 😅 I have no idea how to find it out. `/pull` requests are sent in time shorter than 1 minute (they're pinging active status too). Have you got some ideas?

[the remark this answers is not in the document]
madsmtm: My best guess would be to observe `"buddylist_overlay"` responses while someone becomes inactive? And then see how `"chatproxy-presence"` acts?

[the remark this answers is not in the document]
madsmtm: But I don't really know either :grin:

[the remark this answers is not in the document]
kapi2289: I see in `chatproxy-presence` that there's a `p` attribute that has only `2` or `0` like `a` in buddylist overlay. I think that it is the same (`2` - active, `0` - not active)
```json
"xxxxxxxxxxxx": {
    "lat":1546698870,
    "p":0,
    "vc":74
}
```

fbchat/client.py
kapi2289: Should we raise an error if it isn't?

Issue #335, opened 2018-09-21
KeyError in fetchThreadList()


2FWAH:
## Description of the problem
Fetching threads using fetchThreadList() failed with KeyError.

Method works if no group threads are in ThreadLocation (given as parameter, default to INBOX) but fails if there is at least one group thread (may be related if the user is admin or not of the thread conversation).

## Code to reproduce
Try to fetch threads using fetchThreadList().

## Traceback
```py
>>> client.fetchThreadList()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fb/fbchat/fbchat/client.py", line 895, in fetchThreadList
    return [graphql_to_thread(node) for node in j['viewer']['message_threads']['nodes']]
  File "/home/fb/fbchat/fbchat/client.py", line 895, in <listcomp>
    return [graphql_to_thread(node) for node in j['viewer']['message_threads']['nodes']]
  File "/home/fb/fbchat/fbchat/graphql.py", line 246, in graphql_to_thread
    return graphql_to_group(thread)
  File "/home/fb/fbchat/fbchat/graphql.py", line 306, in graphql_to_group
    approval_requests = set(node["requester"]['id'] for node in group['group_approval_queue']['nodes']),
KeyError: 'group_approval_queue'
```

## Environment information
- Python 3.5.3
- **Master branch** (08117e7a54a33f25812f9bd970e3c2e9cb96a17c)

Looks related to the recent support for rooms.

kapi2289:
Ohh, that's my fault, I'll fix this!

madsmtm:
Closed by #340, I'll release a new version later today ;)

PR #340, opened 2018-09-24
[Fix] `fetchThreadList` fix

kapi2289:
## Description
As in #335 `fetchThreadList` throwed error when in the `ThreadLocation` there was an group. Now it's fixed by this PR.


madsmtm:
👍

PR #342, opened 2018-09-24
[Feature] Quick replies

kapi2289:
## Description
Added
- New `QuickReply` model
- `quick_replies` attribute of `Message` model
- Sending quick replies with `send` method

## Examples
You can send quick replies with text only:
```py
message = Message(text="Do you like me?", quick_replies=[QuickReply(title="Yes!"), QuickReply(title="No")])
client.send(message)
```
Example
or with images:
```py
message = Message(text="Do you like me?", quick_replies=[QuickReply(title="Yes!", image_url="https://example.com/image1.jpg"), QuickReply(title="No", image_url="https://example.com/image2.jpg")])
client.send(message)
```
Example

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -1045,6 +1045,18 @@ class Client(object):
         if message.sticker:
             data['sticker_id'] = message.sticker.uid
 
+        if message.quick_replies:
+            xmd = {"quick_replies": []}
+            for quick_reply in message.quick_replies:
+                xmd["quick_replies"].append({
+                    "content_type": "text",
+                    "title": quick_reply.title,
+                    "payload": quick_reply.payload,
+                    "image_url": quick_reply.image_url,
+                    "data": quick_reply.data,
+                })
+            data['platform_xmd'] = json.dumps(xmd)
+
         return data
 
     def _doSendRequest(self, data, get_thread_id=False):
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -193,6 +193,16 @@ def graphql_to_plan(a):
     rtn.invited = [m.get('node').get('id') for m in guests if m.get('guest_list_state') == "INVITED"]
     return rtn
 
+def graphql_to_quick_reply(q, is_response=None):
+    rtn = QuickReply(
+        title=q.get('title'),
+        image_url=q.get('image_url'),
+        payload=q.get('payload'),
+        data=q.get('data'),
+    )
+    rtn.is_response = is_response
+    return rtn
+
 def graphql_to_message(message):
     if message.get('message_sender') is None:
         message['message_sender'] = {}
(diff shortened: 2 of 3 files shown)


madsmtm:
Looks amazing (!!!), you talked about adding special quick replies, that asks for the users phone number, email, or location, and I think you should add that first (even though it doesn't work on mobile (yet 😉 ))

I think it'd be nice to have a helper method, to reply to a quick reply

Also, the text you wrote above would be nice to have in the examples

fbchat/models.py
```diff
@@ -526,6 +591,12 @@ def __repr__(self):
         # For documentation:
         return '{}.{}'.format(type(self).__name__, self.name)
 
+class QuickReplyType(Enum):
```
madsmtm: There's not really a need for this model any longer

fbchat/graphql.py
```diff
@@ -193,6 +193,24 @@ def graphql_to_plan(a):
     rtn.invited = [m.get('node').get('id') for m in guests if m.get('guest_list_state') == "INVITED"]
     return rtn
 
+def graphql_to_quick_reply(q, is_response=False):
+    data = dict()
+    if q.get('title') is not None: data["title"] = q["title"]
+    if q.get('image_url') is not None: data["image_url"] = q["image_url"]
+    if q.get('payload') is not None: data["payload"] = q["payload"]
+    if q.get('data') is not None: data["data"] = q["data"]
```
madsmtm: Getting the `payload` and `data` attributes outside like this is fine, since they're attributes on the `QuickReply` model, but `title` is only relevant on `QuickReplyText`, so it should only be added there

fbchat/models.py
kapi2289: It's needed because of
```python
q["content_type"] = quick_reply.type.value.lower()
```
in `_getSendData`

fbchat/client.py
```diff
@@ -1141,7 +1141,8 @@ def quickReply(self, quick_reply, thread_id=None, thread_type=None):
         :raises: FBchatException if request failed
         """
         quick_reply.is_response = True
-        return self.send(Message(text=quick_reply.title, quick_replies=[quick_reply]))
+        if quick_reply.type == QuickReplyType.TEXT:
+            return self.send(Message(text=quick_reply.title, quick_replies=[quick_reply]))
```
madsmtm: Why do we only allow textual quick replies? Can't we use the other ones as well?

madsmtm: In any case, this behaviour should be documented instead of being an `if` statement

kapi2289: For location quick reply I need a `sendLocation` method from #333, so I'll wait for merging these two PRs, and for email and phone quick replies I'll make methods to get them and commit it to this PR.

madsmtm: Okay, that makes sense. We should merge #333 first then, and update the implementation before we merge this

fbchat/client.py
```diff
@@ -1040,6 +1064,22 @@ def _getSendData(self, message=None, thread_id=None, thread_type=ThreadType.USER
...
+        if message.quick_replies:
+            xmd = {"quick_replies": []}
+            for quick_reply in message.quick_replies:
+                q = dict()
+                q["content_type"] = quick_reply._type
+                q["payload"] = quick_reply.payload
+                q["external_payload"] = quick_reply.external_payload
+                q["data"] = quick_reply.data
+                if quick_reply.is_response: q["ignore_for_webhook"] = False
+                if quick_reply._type == QuickReplyText._type: q["title"] = quick_reply.title
```
madsmtm:
```suggestion
                if isinstance(quick_reply, QuickReplyText): q["title"] = quick_reply.title
```

fbchat/client.py
```diff
@@ -1040,6 +1064,22 @@ def _getSendData(self, message=None, thread_id=None, thread_type=ThreadType.USER
...
+                if quick_reply._type is not QuickReplyLocation._type: q["image_url"] = quick_reply.image_url
```
madsmtm:
```suggestion
                if not isinstance(quick_reply, QuickReplyLocation): q["image_url"] = quick_reply.image_url
```

madsmtm:
I went ahead and fixed the small issues, thanks a lot for your hard work! 🎉

PR #355, opened 2018-10-27
Allowed empty login when session is given

ekohilas:
Changed login logic to allow empty login details when a session cookie is given.
A slight ease of use and security adjustment such that a password is not required when a session is given.

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -31,7 +31,7 @@ class Client(object):
     Note: Modifying this results in undefined behaviour
     """
 
-    def __init__(self, email, password, user_agent=None, max_tries=5, session_cookies=None, logging_level=logging.INFO):
+    def __init__(self, email=None, password=None, user_agent=None, max_tries=5, session_cookies=None, logging_level=logging.INFO):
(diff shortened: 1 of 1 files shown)


madsmtm:
Well, couldn't you just input `None` manually when you don't want to store the password? I think the default case should be that you supply your credentials together with the session, since that will allow automatic login in if the session was invalid

ekohilas:
Yeah I was thinking about that too... I guess I just like default arguments? :stuck_out_tongue: 

Would you be opposed to adding a login prompt if the login details are none when the session doesn't work?

madsmtm:
Hmm, that'd break for people that are passing `None`s, but it'd be nice for newcomers. I'm not really sure, maybe make an issue about it, asking for other opinions? In any case, the login prompt should be customizable in an `on_` method

madsmtm:
I'll hold off on this for now, I'd guess it's easier for newcomers when they're forced to supply login credentials.

And besides, logging in and loading from a session should ideally have been two different `classmethod`s.

Issue #365, opened 2018-12-07
searchForUsers('<name>') not working as Expected


kaushalvivek:
## Description of the problem
```searchForUsers()``` should return a list of objects that match the name entered as the query, but it's returning only one response *(presumably, the first response.)*

## Code to reproduce
```py
users = client.searchForUsers('Jane')
for user in users:
    print("User's ID: {}".format(user.uid))
```
This should print the user ID of all chats with 'Jane' in their name, but it would return just one option.

## Environment information
- Python 3.7.0
- `fbchat` 1.4.1

kapi2289:
Maybe you have only one Jane on your Facebook friends list?

kaushalvivek:
Oh, no, I checked with multiple search terms. Terms that should definitely have returned multiple results.

kaushalvivek:
https://github.com/carpedm20/fbchat/blob/89a277c3541e5b5a5a1b4c1dd6aaf9025ce43707/fbchat/client.py#L505

Found the issue, the limit is set as 1 here. I suppose that was done for a reason?
5 or 10 would be a better limit logically, I'd say.

madsmtm:
I looked at the history, but I couldn't find an explicit reason for setting the limit so low, maybe it was just to be pragmatic, and encourage people to set their own limits?

kaushalvivek:
It does compromise on usability in its present implementation though, maybe a limit of 3 or 5 would be a better balance between usability and a pragmatic approach?

madsmtm:
Well, you've set it at 10 in #366, which is definitively a better choice for `searchForThreads`, and I don't think the limits should differ between the `searchForX` methods.

I'd still think that the overly pragmatic choice is better. I'd rather add explicit documentation about this fact, than set an arbitary limit. You can see an example of this in the [`asks` documentation](https://asks.readthedocs.io/en/latest/a-look-at-sessions.html#important-connection-un-limiting)

But then again, `searchForMessageIDs`, `searchForMessages` and `search` limits of `5`, and `fetchThreadMessages` and `fetchThreadList` have limits of `20`, so maybe it's not a big deal.

I guess I'll approve this change, and merge the PR when it's updated 👍

kaushalvivek:
Hi @madsmtm , I've updated ```searchForX``` methods for ```Users, Pages, Groups and Threads``` and set the limit to 10. Checkout #366.

PR #366, opened 2018-12-07
Fix for issue #365

kaushalvivek:
Instead of keeping the results for a search query limited to 1, increased it to 10.

Line changed:
https://github.com/carpedm20/fbchat/blob/89a277c3541e5b5a5a1b4c1dd6aaf9025ce43707/fbchat/client.py#L505

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -517,7 +517,7 @@ class Client(object):
 
         return [graphql_to_user(node) for node in j[name]['users']['nodes']]
 
-    def searchForPages(self, name, limit=1):
+    def searchForPages(self, name, limit=10):
         """
         Find and get page by its name
 
@@ -531,7 +531,7 @@ class Client(object):
 
         return [graphql_to_page(node) for node in j[name]['pages']['nodes']]
 
-    def searchForGroups(self, name, limit=1):
+    def searchForGroups(self, name, limit=10):
         """
         Find and get group thread by its name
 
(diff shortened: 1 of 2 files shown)


madsmtm:
Thanks for your contribution, and don't mind the broken tests.

Would it make sense to increase the `limit` on `searchForPages`, `searchForGroups` and `searchForThreads` equivalently?

I looked at the history, but I couldn't find a reason for setting the limit so low, maybe it was just to be pragmatic, and encourage people to set their own limits?

kaushalvivek:
@madsmtm , I would say yeah. Two people/threads/pages might have the same name *(as is for a lot of Indian names :P)*, and this does render searching futile when only one object is being returned. A limit of 3, 5 or 10, would be the best balance between being pragmatic and maintaining usability at the same time. That's just my opinion.

madsmtm:
This would technically be a breaking change if people were doing something like:
```py
try:
    user, = client.searchForUsers("abc")
except:
    pass  # Assume the user couldn't be found
```
But that's bad usage of the method anyhow, so I don't think it'll be a problem.

I'll merge this once you've updated the limit on `searchForPages`, `searchForGroups` and `searchForThreads` as well :)

madsmtm:
Thanks 👍

AuroraRin:
Emergency !!!!! Can any one help me out with the sendLocalVoiceClip, when I run the code it keeps saying that:
'FacebookMess' object has no attribute 'sendLocalVoiceClips'
Ps: My fbchat is 1.6.3. Thanks in advance~

madsmtm:
Hi @AuroraRin, I'm sorry to hear that you're having trouble, and would love to help you out! Unfortunately, this is not the correct place to discuss the issue. Please create a dedicated [bug report](https://github.com/carpedm20/fbchat/issues/new?template=bug_report.md), and attempt to fill in as much information as you can, then I'll get back to you 😊

Issue #302, opened 2018-06-19
Encoding characters


klepretr:
Hi !
I would like to test today but I have an issue and I don't know where I can find a solution.
Do you know a solution for this ?

Thanks in advance

package version: 
```
pip show fbchat
Name: fbchat
Version: 1.3.8
Summary: Facebook Chat (Messenger) for Python
Home-page: https://github.com/carpedm20/fbchat/
Author: Taehoon Kim; Moreels Pieter-Jan; Mads Marquart
Author-email: carpedm20@gmail.com
License: BSD License
Location: /usr/local/lib/python2.7/dist-packages
Requires: enum34, requests, lxml, beautifulsoup4
Required-by:
```

my test script `script.py`:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from fbchat import Client

client = Client('xxxxx', 'xxxx')
print client.fetchAllUsers()
```

execution result:
```
python script.py
Logging in xxxxx...
Login of xxxxx successful.
[Traceback (most recent call last):
  File "script.py", line 8, in <module>
    print client.fetchAllUsers()
UnicodeEncodeError: 'ascii' codec can't encode character u'\xea' in position 18: ordinal not in range(128)
```

madsmtm:
Hmm, it looks like you have an invisible character inside the text `fetchAllUsers`? Try removing it, and writing it again?

klep24:
I have exactly the same issue when I use my terminal
```
$ python
Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from fbchat import Client
>>> client = Client('xxxxx', 'xxxxx')
Logging in xxxxx...
Login of xxxxx successful.
>>> print client.fetchAllUsers()
[Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xea' in position 18: ordinal not in range(128)
```

To be sure this is a issue with encoding, I use `print client.searchForUsers('bastien')` because I know there are not characters with accents or what ever in this name. And as we can see, there are no problems.
```
python script.py  
Logging in xxxxx...
Login of xxxxx successful.
[<USER Bastien (xxxxx)>]
```

Maybe an issue with bad version of python or dependencies.

madsmtm:
[...]

PR #371, opened 2018-12-12
Fix enums

### LINKED ISSUE #220: ThreadColor Enums Error

### LINKED ISSUE #368: ValueError: '🎄' is not a valid MessageReaction
## Description of the problem
My friends somehow (probably with the android client which for some reason has more features / is less limited) managed to react with a 🎄. The result is that I can't fetch messages up to a certain point (or well, not directly at least).
Other emojis work as well, of course.
schermafbeelding 2018-12-12 om 12 54 51
schermafbeelding 2018-12-12 om 12 54 56
schermafbeelding 2018-12-12 om 12 55 01

## Code to reproduce
```py
from fbchat import Client
client = Client("<email>", "<password>")
client.fetchThreadMessages('1234567890', limit=400)
```

## Traceback
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/fbchat/client.py", line 847, in fetchThreadMessages
    messages = list(reversed([graphql_to_message(message) for message in j['message_thread']['messages']['nodes']]))
  File "/usr/local/lib/python3.7/site-packages/fbchat/client.py", line 847, in <listcomp>
    messages = list(reversed([graphql_to_message(message) for message in j['message_thread']['messages']['nodes']]))
  File "/usr/local/lib/python3.7/site-packages/fbchat/graphql.py", line 212, in graphql_to_message
    rtn.reactions = {str(r['user']['id']):MessageReaction(r['reaction']) for r in message.get('message_reactions')}
  File "/usr/local/lib/python3.7/site-packages/fbchat/graphql.py", line 212, in <dictcomp>
    rtn.reactions = {str(r['user']['id']):MessageReaction(r['reaction']) for r in message.get('message_reactions')}
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 309, in __call__
    return cls.__new__(cls, value)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 543, in __new__
    return cls._missing_(value)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 556, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: '🎄' is not a valid MessageReaction
```

## Environment information
- Python version: 3.7.1
- `fbchat` version: 1.4.2

madsmtm:
Should fix #220 and fix #368. `changeThreadColor` and `reactToMessage` are not aware of this change yet.

I'm having trouble with Python 2.7 encoding. Narrowed it down to:
```py
tree = '🎄'
u"abc_{}".format(tree)
```
If we can figure the encoding errors, we might be able to solve #302?

### DIFF
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -214,7 +214,16 @@ def graphql_to_message(message):
     rtn.timestamp = message.get('timestamp_precise')
     if message.get('unread') is not None:
         rtn.is_read = not message['unread']
-    rtn.reactions = {str(r['user']['id']):MessageReaction(r['reaction']) for r in message.get('message_reactions')}
+
+    for r in message.get('message_reactions'):
+        try:
+            reaction = MessageReaction(r['reaction'])
+        except ValueError:
+            log.warning("Failed parsing reaction {}. Extending enum.".format(r['reaction']))
+            aenum.extend_enum(MessageReaction, "UNKNOWN_{}".format(r['reaction']), r['reaction'])
+            reaction = MessageReaction(r['reaction'])
+        rtn.reactions[str(r['user']['id'])] = reaction
+
     if message.get('blob_attachments') is not None:
         rtn.attachments = [graphql_to_attachment(attachment) for attachment in message['blob_attachments']]
     # TODO: This is still missing parsing:
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 import json
 import re
+import aenum
 from .models import *
 from .utils import *
 
@@ -26,12 +27,16 @@ class ConcatJSONDecoder(json.JSONDecoder):
 def graphql_color_to_enum(color):
     if color is None:
         return None
-    if len(color) == 0:
+    if not color:
         return ThreadColor.MESSENGER_BLUE
+    color = color[2:]  # Strip the alpha value
+    color_value = '#{}'.format(color.lower())
     try:
-        return ThreadColor('#{}'.format(color[2:].lower()))
+        return ThreadColor(color_value)
     except ValueError:
-        raise FBchatException('Could not get ThreadColor from color: {}'.format(color))
+        log.warning("Failed parsing color {}. Extending enum.".format(color))
+        aenum.extend_enum(ThreadColor, "UNKNOWN_{}".format(color), color_value)
+        return ThreadColor(color_value)
 
 def get_customization_info(thread):
     if thread is None or thread.get('customization_info') is None:
--- a/fbchat/graphql.py
+++ b/fbchat/graphql.py
@@ -24,6 +24,14 @@ class ConcatJSONDecoder(json.JSONDecoder):
         return objs
 # End shameless copy
 
+def enum_extend_if_invalid(enumeration, value):
+    try:
+        return enumeration(value)
+    except ValueError:
+        log.warning("Failed parsing {}({!r}). Extending enum.".format(enumeration, value))
+        aenum.extend_enum(enumeration, "UNKNOWN_{}".format(value).upper(), value)
+        return enumeration(value)
+
 def graphql_color_to_enum(color):
     if color is None:
         return None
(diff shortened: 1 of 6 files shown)


fbchat/graphql.py
```diff
@@ -26,12 +26,11 @@ def decode(self, s, _w=WHITESPACE.match):
 def graphql_color_to_enum(color):
     if color is None:
         return None
-    if len(color) == 0:
+    if not color:
         return ThreadColor.MESSENGER_BLUE
```
Demindiro: I think this line won't execute because there is a None check before this one (or is the `not color` intended to catch another value?)

madsmtm: The `not color` is intended to catch the empty string, but I guess that could be clearer

madsmtm:
I'll go ahead and merge this, I still don't really know what to do about the encoding errors yet, though.

coolchip25:
I might have botched some syntax, but I'm having issues with the `client.fetchThreadList()` function.

After signing in if I run 
`threads = client.fetchThreadList()`

I get the following:

```
Traceback (most recent call last):
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\graphql.py", line 32, in graphql_color_to_enum
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\enum.py", line 309, in __call__
    return cls.__new__(cls, value)
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\enum.py", line 545, in __new__
    return cls._missing_(value)
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\enum.py", line 558, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: '#f01d6a' is not a valid ThreadColor

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\client.py", line 903, in fetchThreadList
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\client.py", line 903, in <listcomp>
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\graphql.py", line 246, in graphql_to_thread
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\graphql.py", line 291, in graphql_to_group
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\graphql.py", line 43, in get_customization_info
  File "C:\Users\curti\AppData\Local\Programs\Python\Python37-32\lib\site-packages\fbchat\graphql.py", line 34, in graphql_color_to_enum
fbchat.models.FBchatException: Could not get ThreadColor from color: FFF01D6A
```
I used to use this code all the time and it worked fine. Might also be worth mentioning that it always fails on the first login attempt for me (but succeeds on the second). Also I am on version 1.6.5

PR #378, opened 2019-01-06
Added OnWaved

ThatAlexanderA:
I'm using lwa_type to detect the waveing. I'm not sure if this is the most efficent way to get lwa_type.

Maybe if we get the lwa_type here, is a better way?
```
elif mercury.get('extensible_attachment'):
      attachment = graphql_to_extensible_attachment(mercury['extensible_attachment'])
      if isinstance(attachment, UnsentMessage):
            unsent = True
             elif attachment:
                attachments.append(attachment)
```


fbchat/client.py
```diff
@@ -2106,6 +2106,15 @@ def getThreadIdAndThreadType(msg_metadata):
                     delta = m["delta"]
                     delta_type = delta.get("type")
                     delta_class = delta.get("class")
+                    
+                    delta_attachments = delta.get('attachments')
+                    lwa_type = None
+
+                    for a in delta_attachments:
```
kapi2289: Going through attachments is already done in [listening for messages](https://github.com/carpedm20/fbchat/blob/3e7b20c379d552944046ab2f3224b9e7903c9d7f/fbchat/client.py#L2386)

fbchat/client.py
```diff
@@ -2260,6 +2269,11 @@ def getThreadIdAndThreadType(msg_metadata):
                         self.onUserJoinedCall(mid=mid, joined_id=author_id, is_video_call=is_video_call,
                                             thread_id=thread_id, thread_type=thread_type, ts=ts, metadata=metadata, msg=m)
 
+                    # On Waved to thread
+                    elif lwa_type == "WAVE":
```
kapi2289: `onWaved` will be called, but `onMessage` not, although waving is a message too

fbchat/client.py
```diff
@@ -3035,6 +3049,18 @@ def onUserJoinedCall(self, mid=None, joined_id=None, is_video_call=None, thread_
         """
         log.info("{} joined call in {} ({})".format(joined_id, thread_id, thread_type.name))
 
+        
+        
+    def onWaved(self, author_id, thread_id=None, thread_type=None, ts=None):
```
kapi2289: For future, `author_id` should be `None`, and by the way you should include `msg` and `metadata` parameters too (like in other methods)

ThatAlexanderA:
Okay, sorry for the very late reply I was very busy. I will add the msg and metadata to the def now. And at the weekend I try to make the WaveAttachment.

kapi2289:
And remember to use the `black` linter! Look at the `CONTRIBUTING.rst` guide :wink:

TheLastGimbus:
Thank you, I was just looking forward for that feature!

kapi2289:
It's been a long time and the change requests haven't been applied. Closing.

Issue #392, opened 2019-02-10
MessageLocation breaks fetchThreadMessages()


darylkell:
## Description of the problem
Thanks for the great library!

Was working swimmingly until I added a MessageLocation to the test thread. Now provides the error below as it fetchThreadMessages(thread).
I have not tested a MessageLiveLocation in the test thread.

The particular location it failed on (url variable in graphql_to_extensible_attachment()) is:
https://l.facebook.com/l.php?u=https%3A%2F%2Fwww.bing.com%2Fmaps%2Fdefault.aspx%3Fv%3D2%26pc%3DFACEBK%26mid%3D8100%26where1%3DHelena%2BValley%252C%2BPerth%252C%2BWestern%2BAustralia%2B6155%26FORM%3DFBKPL1%26mkt%3Den-GB&h=AT1XzJ_ST9Xlk863CIHbJ05kI6BdeHPyfPJf_keuBNjFZR4iqKAQIL8DBqVKiGNmZHD3xnUkj4qVASGOD_2vyu9zifMRGN6nJvCFNSjsUAUvtxcgNx5DWq6uuqviewuP&s=1

## Traceback
```
Traceback (most recent call last):
  File "C:/Users/D-PC/Google Drive/Work/Projects/99 - Other/FBchat/Facebook_message_download-master/Message_windows.py", line 307, in <module>
    do_rest(thread)
  File "C:/Users/D-PC/Google Drive/Work/Projects/99 - Other/FBchat/Facebook_message_download-master/Message_windows.py", line 146, in do_rest
    messages = client.fetchThreadMessages(thread_id=thread_id, limit=1000, before=timestamp)
  File "C:\Program Files\Python370\lib\site-packages\fbchat\client.py", line 1105, in fetchThreadMessages
    for message in j["message_thread"]["messages"]["nodes"]
  File "C:\Program Files\Python370\lib\site-packages\fbchat\client.py", line 1105, in <listcomp>
    for message in j["message_thread"]["messages"]["nodes"]
  File "C:\Program Files\Python370\lib\site-packages\fbchat\graphql.py", line 395, in graphql_to_message
    attachment = graphql_to_extensible_attachment(message["extensible_attachment"])
  File "C:\Program Files\Python370\lib\site-packages\fbchat\graphql.py", line 148, in graphql_to_extensible_attachment
    ).split(", ")
ValueError: too many values to unpack (expected 2)

```

## Environment information
- Python version 3.7.1
[...]

PR #395, opened 2019-02-14
LocationAttachment fix

darylkell:
Fix for locations where address is provided instead of GPS co-ords (issue #392)

- Add self.location attribute to class LocationAttachment to house address string
- Fix for graphql_to_extensible_attachment(a) to include location attribute for addresses, while maintaining current lat/long attributes


darylkell:
I can see the black lint check failed, and I have seen on a previous PR that this is requested to be fixed by referring to 'CONTRIBUTING.rst' which to be honest is not a guide at all.

As mentioned 20 days ago by madsmtm in a PR, the use of flint has indeed increased the initial difficulty of manual installation (users have to learn a new tool) and unfortunately I can't help any more than provide what I have in this initial PR.

Anyway, I hope it helps.

kapi2289:
Please run this in your `fbchat` directory
```console
$ pip install -U black
$ black .
```
I would change the `location` attribute name to `address`.
And please mention in the attribute docs that available are only the coordinates OR only the address

fbchat/models.py
```diff
@@ -456,12 +458,15 @@ class LocationAttachment(Attachment):
     image_height = None
     #: URL to Bing maps with the location
     url = None
+    # Address of the location
+    address = None
 
-    def __init__(self, latitude=None, longitude=None, **kwargs):
+    def __init__(self, latitude=None, longitude=None, address=None, **kwargs):
```
madsmtm: I think `latitude` and `longitude` are available in the constructor because they have a relevance if you send the location using `Client.sendLocation` or likewise, but there's no reason for ever setting `address` yourself.

But this is super nitpicky, and the described pattern isn't consistent throughout the rest of the code, so it probably doesn't really matter 😁

fbchat/graphql.py
```diff
@@ -143,13 +143,20 @@ def graphql_to_extensible_attachment(a):
...
                 url = story.get("url")
-                latitude, longitude = get_url_parameter(
-                    get_url_parameter(url, "u"), "where1"
-                ).split(", ")
+                address = get_url_parameter(get_url_parameter(url, "u"), "where1")
+                if len(address.split(", ")) == 2:
+                    try:
+                        latitude, longitude = [float(x) for x in address.split(", ")]
+                        address = ""
+                    except:
```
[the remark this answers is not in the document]
darylkell: That would make sense.

fbchat/graphql.py
```diff
@@ -143,13 +143,20 @@ def graphql_to_extensible_attachment(a):
             _type = target["__typename"]
             if _type == "MessageLocation":
                 url = story.get("url")
-                latitude, longitude = get_url_parameter(
-                    get_url_parameter(url, "u"), "where1"
-                ).split(", ")
+                address = get_url_parameter(get_url_parameter(url, "u"), "where1")
+                if len(address.split(", ")) == 2:
+                    try:
+                        latitude, longitude = [float(x) for x in address.split(", ")]
```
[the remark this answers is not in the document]
darylkell: Apart from storing address.split() out to a variable and then calling on that variable twice instead, I can't think of a way to reduce the second split() method-call overhead. I was going for less lines, less variables, more readability, at the cost of the second split() call.

fbchat/graphql.py
```diff
@@ -143,13 +143,20 @@ def graphql_to_extensible_attachment(a):
...
+                        address = ""
+                    except:
+                        latitude, longitude = "", ""
+                else:
+                    latitude, longitude = "", ""
```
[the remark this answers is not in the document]
darylkell: Fair call.

fbchat/models.py
darylkell: I figured if some message locations only have addresses and no co-ords that it must mean they're also able to be constructed in this way (by sending a recogniseable address, not just co-ords) so I wanted to account for that. If it's not the case, then what you're saying makes sense. I just don't know why Facebook would then return a location without co-ords, but only the address.

darylkell:
I'm happy for you to do whatever you like with the code I've presented to make it viable for your project. I've had a crack at using GitHub to fork/push request etc and learned a bit trying to contribute to FBchat - and I appreciate you bearing with me along the way. 
That said, I'm not bothered by having a successful PR contribution next to my name, just wanted to offer a means to improve the API so perhaps you can find some use from what I've added here.

fbchat/graphql.py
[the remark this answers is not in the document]
kapi2289: You can make it like it was made, and just use move this into `try`
```python 
try:
    latitude, longitude = address.split(", ")
    (...)
```

fbchat/models.py
madsmtm: Oh yeah, that does make sense. Perhaps the task is then to update `Client.sendLocation` instead. But that'll probably be really difficult to figure out

darylkell: Maybe not 'instead', maybe 'as well as'. This change doesn't break client.sendLocation - but it does fix the receipt of messages with locations that don't have co-ords. 
client.sendLocation still functions, just limited to sending co-ords. Adding functionality to send an address (that Facebook accepts) seems a separate issue to me... :)

madsmtm: Yeah, sorry, I meant "as well as" 😉. And yeah, it's a separate issue!

PR #400, opened 2019-02-24
Move pull delta parsing into separate method

madsmtm:
A pure refactor, no actual code changes. Reduces the size of `_parseMessage` and makes the code more readable (less indentation).

Made as two commits, since `git` would get really confused otherwise. It's still hard to review, I know 😕


kapi2289:
It would be nice too to move all these parsing methods into some other class in other file, like `_listening.py`, and inherit this class in the `Client`

madsmtm:
Indeed! This was just the initial push towards something a little bit more readable 😉

Issue #396, opened 2019-02-18
Version 2 wishlist


madsmtm:
I've been brewing ideas for `v2` for a while now (probably over a year), and I started implementing some of these ideas in a separate branch (completely rewriting the code). Inevitably, the `master` branch kept getting worked on, and it eventually fell apart.

So now I'm trying to reboot some of my ideas, but I'll take a separate approach to implementing them: I'll implement almost everything that can be done without making breaking changes, and then I'll move the project into a state of `v2`, where we can properly progress towards a cleaner API. In general, I'd much rather have a stable, minimal API, than an unstable, comprehensive API!

I've tried to assemble a list of major things I'd like to see before making breaking changes:

- [x] Use `flit` as our build system (see #382)
- [x] Use `black` for formatting (see #386)
- [ ] Refactor code into logical, separate files (`client.py` is far too large)
  - [ ] Would help us to make proper unit tests
- [ ] Set up integration tests using something like [VCR.py](https://github.com/kevin1024/vcrpy)
- [ ] Clean up documentation
  - [ ] Fix setup (tracked in #391)
  - [ ] API rundown (Use [Google Docstring Style](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings))
  - [ ] Introduction overview overhaul
  - [ ] Investigate tools for detecting docstring errors and spelling mistakes
- [ ] Inline examples
- [ ] More detailed examples (recommends good practice too)

And a list of major things I'd like to change:

- [ ] Declare a proper public API. See [trio](https://github.com/python-trio/trio) as a reference to a project that fullfulls this promise (e.g. using `_` before methods and filenames)
- [ ] Declare a concise API (e.g. document expected exceptions)
  - [ ] Using datetimes, see #278, is a subtask of this
- [ ] Use camel_cased method names like PEP8 suggests
- [ ] Fix logging (see #258)
- [ ] Make a distinction between models, like `Message`, and what you can send, as parameters in `send`. Though combining these sound good idea in theory, it's really messy in practice, and not really that useful.
- [ ] Make a distinction between user created models and recieved models
  - [ ] Building upon this idea, we can move a lot of implementations into the model instead of a central client. For example, I'd like to be able to do something like:
    ```py
    message = Message(id="123", client=client)  # Does not make an external call (since you don't need to, to interact with the message)
    message.react(None)
    assert not hasattr(message, "text")
    # Or alternatively:
[...]

PR #401, opened 2019-02-27
Create multi-threading.py (example)

joakimed:
New example for simultaneous sending and receiving, as discussed in the last few comments on #396 .
I'm sure there are better ways of doing this, so please hit me with any ideas for improvement. 

Also, the script is a little buggy when it comes to printing. For instance; if you are in the middle of writing something and you receive another message, it gets printed in the middle of what you are writing. I was unable to find a solution without importing a third-party library.

### DIFF
--- a/examples/multi-threading.py
+++ b/examples/multi-threading.py
@@ -1,0 +1,42 @@
+# -*- coding: UTF-8 -*-
+
+from fbchat import Client
+from fbchat.models import *
+import threading
+
+# Subclass fbchat.Client and override required methods
+class PrintMessage(Client):
+    def onMessage(self, author_id, message_object, thread_id, thread_type, **kwargs):
+        self.markAsDelivered(thread_id, message_object.uid)
+        self.markAsRead(thread_id)
+
+        print("\nIncoming message: {}".format(message_object.text))
+
+    # Creating some feedback on login since we will disable INFO logging
+    def onLoggedIn(self, email=None):
+        print("Login of {} successful.".format(email))
+
+
+# Logging in and setting logging level to WARNING to avoid some unessential output
+client = PrintMessage("<email>", "<password>", logging_level=30)
+
+
+def send():
+    while True:
+        payload = input("Message: ")
+        if payload:
+            client.send(
+                Message(text=payload), thread_id=client.uid, thread_type=ThreadType.USER
+            )
+
+
+def receive():
+    while True:
+        client.doOneListen()
+
+
+# Creating and starting separate threads for handling the receiving and sending of messages
+t1 = threading.Thread(target=receive)
+t2 = threading.Thread(target=send)
+t1.start()
+t2.start()


examples/multi-threading.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: UTF-8 -*-
+
+from fbchat import Client
+from fbchat.models import *
+import threading
+
+# Subclass fbchat.Client and override required methods
+class PrintMessage(Client):
```
madsmtm: It seems wrong to call the client `PrintMessage`. Maybe something like `MessagePrinter` or `PrintBot` is better?

examples/multi-threading.py
```diff
@@ -0,0 +1,42 @@
...
+        payload = input("Message: ")
+        if payload:
+            client.send(
+                Message(text=payload), thread_id=client.uid, thread_type=ThreadType.USER
+            )
+
+
+def receive():
+    while True:
+        client.doOneListen()
```
madsmtm: This is wrong, you need to call `client.startListen` and `client.stopListen` as well. See `Client.listen` for how this is supposed to work. But really, is there a reason to not just use `client.listen` as the `target` in `threading.Thread`?

examples/multi-threading.py
```diff
@@ -0,0 +1,42 @@
...
+
+# Logging in and setting logging level to WARNING to avoid some unessential output
+client = PrintMessage("<email>", "<password>", logging_level=30)
+
+
+def send():
+    while True:
+        payload = input("Message: ")
+        if payload:
+            client.send(
```
madsmtm: Again, I don't think operating the listener and sending messages on the same client is thread safe! At least not formally. So I'm not sure if it's something we wanna publicly recommend?

examples/multi-threading.py
```diff
@@ -0,0 +1,42 @@
...
+def receive():
+    while True:
+        client.doOneListen()
+
+
+# Creating and starting separate threads for handling the receiving and sending of messages
+t1 = threading.Thread(target=receive)
+t2 = threading.Thread(target=send)
+t1.start()
+t2.start()
```
madsmtm: Do we need a `join` somewhere? Or is that automatically done?

examples/multi-threading.py
```diff
@@ -0,0 +1,42 @@
...
+    # Creating some feedback on login since we will disable INFO logging
+    def onLoggedIn(self, email=None):
+        print("Login of {} successful.".format(email))
+
+
+# Logging in and setting logging level to WARNING to avoid some unessential output
+client = PrintMessage("<email>", "<password>", logging_level=30)
+
+
+def send():
```
madsmtm: This should probably be the main thread, while the listener is spawned in a child process (Instead of spawning two child processes)

madsmtm: That way you could also easily stop the listener when needed

examples/multi-threading.py
```diff
@@ -0,0 +1,42 @@
...
+
+        print("\nIncoming message: {}".format(message_object.text))
+
+    # Creating some feedback on login since we will disable INFO logging
+    def onLoggedIn(self, email=None):
+        print("Login of {} successful.".format(email))
+
+
+# Logging in and setting logging level to WARNING to avoid some unessential output
+client = PrintMessage("<email>", "<password>", logging_level=30)
```
madsmtm: Instead of `30`, use `logging.WARNING`

examples/multi-threading.py
joakimed: `MessagePrinter` it is.

examples/multi-threading.py
joakimed: I assume you mean it's not safe to send a message while listening? So by stopping the listening thread while sending, we should be OK?

madsmtm: That would probably be enough! But on the other hand, it's really inefficient, and not really clean either 😕. Maybe the best solution is just to create two clients?

madsmtm:
I've been investigating a bit, I'd guess that https://github.com/kennethreitz/requests/issues/2766 would make it possible to solve the thread safety problems, but our code needs to be implemented with thread safety in mind as well, so it might not be that simple!

examples/multi-threading.py
joakimed: I removed one thread, and made the remaining thread daemonic. From what I gather this should get destroyed when the main program exits.

examples/multi-threading.py
```diff
@@ -17,26 +22,29 @@ def onLoggedIn(self, email=None):
...
-
-
-# Creating and starting separate threads for handling the receiving and sending of messages
-t1 = threading.Thread(target=receive)
-t2 = threading.Thread(target=send)
-t1.start()
-t2.start()
+# Clean-up on exit
+except KeyboardInterrupt:
+    client1.logout()
```
madsmtm: I'd argue there's really not a big need to log out (it's not shown in the other examples, either)

examples/multi-threading.py
```diff
@@ -17,26 +22,29 @@ def onLoggedIn(self, email=None):
...
+    client2.logout()
+    sys.exit(0)
```
madsmtm: We should clean the `client2` up properly instead. That can be done by setting `Client.listening` to `False`, and then wait for the thread to clean up afterwards, using `Thread.join`

madsmtm: And then there'd be no need to make `t1` daemonic

PR #403, opened 2019-03-01
Allow to connect to Facebook through TOR

rezemika:
Hi!

Here is an humble proposal of a simple modification that allows to connect to Facebook through TOR, as if a user would use Tor Browser to [facebookcorewwwi.onion](facebookcorewwwi.onion). I wrote this for a personal project, so here is a PR if you think it can be useful. :)

### DIFF
--- a/fbchat/client.py
+++ b/fbchat/client.py
@@ -43,6 +43,7 @@ class Client(object):
         email,
         password,
         user_agent=None,
+        use_tor=False,
         max_tries=5,
         session_cookies=None,
         logging_level=logging.INFO,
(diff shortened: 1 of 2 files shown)


madsmtm:
Thanks for the PR! However, I probably won't merge it, I'm trying to take a more conservative stand on non-essential functionality, at least while I'm doing a lot of code cleanup 😊. It should be possible to provide this functionality by extending `Client` and `ReqUrl` manually, though. If you provided that as an example instead, then I'll probably redecide 😉

rezemika:
Well, I'm sorry for the late of my answer.

I'm not sure it's possible to provide this as an example with changing anything in the base code, because it needs to change `Client` attributes before the end of its `__init__` method. For example, it's not possible to change the `self._session` attribute, but it's mandatory to set the new proxies.

Else, it's possible to create a class which inherit from `Client` and overrides its `__init__` method, but it won't work anymore if you plan to rewrite it and change the way it works in your future code cleanup.

madsmtm:
Yeah, I can see that. I think I'll keep this PR open until we reach `v2`, where we can then discuss how this can be implemented (the code in `__init__` will probably be restructured in such a fasion that this change will be possible)

PR #406, opened 2019-03-07
Refactor model parsing

madsmtm:
Move parsing code out from primarily `_util.py`, `_graphql.py` and `Client._parseMessage` and into the different models.

Breaking changes:
- Remove `graphql.py` (though `graphql_queries_to_json`, `graphql_response_to_json` and `GraphQL` are still exported globally)
- Remove `utils.enum_extend_if_invalid`, `utils.get_emojisize_from_tags`, `utils.LIKES` and `utils.GENDERS`
- `Plan.going`, `Plan.declined` and `Plan.invited` are now shortcut accessors to `Plan.guests`, which means you can't set them anymore! (This change should probably have been done in a separate PR... Sorry 😉).

But all of that was undocumented code/undefined behaviour, so it probably wasn't widely used. I will bump the minor version number before releasing this, though!


madsmtm:
Status: This is fairly done, but since I've just released `v1.6.5`, I'll wait a week or so before merging this into `master` (in case we have to do some hotfixes)

PR #420, opened 2019-04-18
Add forwarding attachments

kapi2289:
```python
client.forwardAttachment("[REDACTED ATTACHMENT ID]", "[REDACTED THREAD ID]")
```


fbchat/_client.py
```diff
@@ -1715,6 +1715,23 @@ def sendLocalImage(
...
+        :param attachment_id: Attachment ID to forward
+        :param thread_id: User/Group ID to send to. See :ref:`intro_threads`
+        :raises: FBchatException if request failed
+        """
+        thread_id, thread_type = self._getThread(thread_id, None)
+        data = {
+            "attachment_id": attachment_id,
+            "recipient_map[{}]".format(generateOfflineThreadingID()): thread_id,
+        }
+        j = self._post(
```
madsmtm: Nothing useful in the response, I take?

kapi2289: Only user IDs, but we don't need it

madsmtm: Just checked, there's at least a `success` / `error` parameter.

Successful:
```
{
    "__ar": 1,
    "payload": {
        "success": true,
        "error": null,
        "user_ids": [
            "[REDACTED]"
        ]
    },
    "bootloadable": {},
    "ixData": {},
    "bxData": {},
    "gkxData": {},
    "qexData": {},
    "lid": "[REDACTED]"
}
```

Failed:
```
{
    "__ar": 1,
    "payload": {
        "success": false,
        "error": "Fanout recipient cannot be the same as the mailbox id",
        "message_targets": [
            "[REDACTED]"
        ]
    },
    "bootloadable": {},
    "ixData": {},
    "bxData": {},
    "gkxData": {},
    "qexData": {},
    "lid": "[REDACTED]"
}
```

So maybe we should extend `check_json` to account for this format as well?

fbchat/utils.py
```diff
@@ -26,6 +26,7 @@
     check_request,
     get_jsmods_require,
     get_emojisize_from_tags,
+    get_forwarded_from_tags,
```
madsmtm: Hmm, I actually made this file so that it would be easy to see when we made backwards incompatible changes, not just to re-export everything from `_util.py`. And I don't think I want to make this function public, since it's mostly an implementation detail.

PR #426, opened 2019-04-27
Add missing mid parameter in message parsing

kapi2289:


madsmtm:
Nice, thanks for the fix! 👍

PR #434, opened 2019-06-30
WIP: fetchThreadImages or feature request

krzesu0:
## Description
I wanted to create a way of downloading all images posted sent in given thread.

# Research
I stareted by watching network traffic on `messenger.com` while scrolling down the image history. \
First query looked like this:
`https://www.messenger.com/webgraphql/query/?query_id=515216185516880&variables={"id":"<uid>","first":"<number>"}`.Where uid is specified user and number is how many to download, default is 12. Facebook will do more, but I am kinda scared.

If you scroll down further, it shows more and more history.\
Query now looks like this:
`https://www.messenger.com/webgraphql/query/?query_id=515216185516880&variables={"id":"<uid>","after":"<image_id>","first":<number>}`. Previous variables are the same, but there is new thing called `after`, after digging around some more I found out that image_id is special variable for starting query after given image. It is called coursor in JSON.

After POST'ing it, it returns almost readable JSON, it begins with "for (;;);".

## My development
I forked it and created my own crude function to handle it. I'll keep working on it.


madsmtm:
Hi there @krzesu0, thanks for the research!

I've previously investigated Facebook's GraphQL endpoints quite extensively, and i know that the endpoint at `/webgraphql/query` works similarly to the one at `/api/graphqlbatch/`, which we've previously implemented support for, see `Client.graphql_request`. So I'd suggest using that, something like this should get you far:
```python
data = {"id": <thread_id>, "first": 12}
j = self.graphql_request(_graphql.from_query_id("515216185516880", data))
```

Note: This internal API was changed recently in #439, so remember to update from `master` before implementing this!

When you want to parse the result, please make a helper `@classmethod` on the `_file.ImageAttachment` class, and convert the result to `ImageAttachment` objects.

Regarding the cursors, then I think it'd be nice if we could hide them as an implementation detail, and just return an iterable, then the user can choose how many results they want from there?

fbchat/_util.py
```diff
@@ -53,6 +53,95 @@
 ]
 
 
+<<<<<<< HEAD
```
kapi2289: You've forgot to remove these ;)

krzesu0:
> When you want to parse the result, please make a helper `@classmethod` on the `_file.ImageAttachment` class, and convert the result to `ImageAttachment` objects.
> 
> Regarding the cursors, then I think it'd be nice if we could hide them as an implementation detail, and just return an iterable, then the user can choose how many results they want from there?

I rewrote this as @madsmtm suggested. I decided to hide cursor from user . Both `_fetchImages` (helper) and `fetchThreadImages` are now generator objects.

fbchat/_util.py
```diff
@@ -53,6 +53,91 @@
 ]
 
 
+class ReqUrl(object):
```
kapi2289: This class was removed in #433, you've just merged your PR incorrectly ;)

krzesu0: Damn it.

fbchat/_file.py
```diff
@@ -155,6 +155,18 @@ def _from_graphql(cls, data):
...
         )
 
+    @classmethod
+    def _from_list(cls, data):
+        data = data["node"]
+        return cls(
+            width=data["original_dimensions"].get("x"),
+            height=data["original_dimensions"].get("y"),
+            thumbnail_url=data["image"].get("uri"),
+            large_preview=data["image2"],
```
madsmtm: Do the previews have a known, constant size? If so, we should provide the sizes here

fbchat/_client.py
```diff
@@ -976,6 +976,42 @@ def getUserActiveStatus(self, user_id):
         """
         return self._buddylist.get(str(user_id))
 
+    def _fetchImages(self, thread_id):
```
madsmtm: Is there a reason for having the two methods? If not, I'd prefer only having one

fbchat/_client.py
```diff
@@ -976,6 +976,42 @@ def getUserActiveStatus(self, user_id):
...
+                yield j[thread_id]["message_shared_media"]["edges"][0]
+                del j[thread_id]["message_shared_media"]["edges"][0]
+            except IndexError:
+                if page_info.get("has_next_page"):
+                    data["after"] = end_cursor
+                    j = self.graphql_request(
+                        _graphql.from_query_id("515216185516880", data)
+                    )
+                else:
+                    to_continue = False
```
madsmtm: Use `break` or `return` instead of keeping track of a variable like this!

fbchat/_client.py
```diff
@@ -976,6 +976,42 @@ def getUserActiveStatus(self, user_id):
         """
         return self._buddylist.get(str(user_id))
 
+    def _fetchImages(self, thread_id):
+        data = {"id": thread_id, "first": 12}
```
madsmtm: Does it make sense to implement some sort of incremental fetching amount? E.g. start by fetching the first 10 images, and the next time 20, and the next time 40, and so on?

fbchat/_client.py
```diff
@@ -976,6 +976,42 @@ def getUserActiveStatus(self, user_id):
...
+                    data["after"] = end_cursor
+                    j = self.graphql_request(
+                        _graphql.from_query_id("515216185516880", data)
+                    )
+                else:
+                    to_continue = False
+
+    def fetchThreadImages(self, thread_id=None):
+        """
+        Creates generator object for fetching images posted in thread.
```
madsmtm: Could you add some example usage to `examples/fetch.py`, perhaps along with usage of [`itertools.islice`](https://docs.python.org/3/library/itertools.html#itertools.islice) to fetch e.g. the first 20 images.

fbchat/_client.py
```diff
@@ -976,6 +976,42 @@ def getUserActiveStatus(self, user_id):
         """
         return self._buddylist.get(str(user_id))
 
+    def _fetchImages(self, thread_id):
+        data = {"id": thread_id, "first": 12}
+        j = self.graphql_request(_graphql.from_query_id("515216185516880", data))
+        to_continue = True
+        while to_continue:
+            page_info = j[thread_id]["message_shared_media"]["page_info"]
+            end_cursor = page_info.get("end_cursor")
```
madsmtm: You should only fetch the `end_cursor` when it's needed (aka. further down)

fbchat/_client.py
krzesu0: I am scared to do more than the messengers default 12. But yeah, I could do that.

krzesu0: My bad! After digging around some more it looks like if you enter "big screen" mode to view images it fetches 48 of them. I'll do 48 items at the time.

fbchat/_client.py
krzesu0: Its easier to manage code this way for me, but yeah, I'll merge them.

krzesu0:
I don't think this feature will work for thread videos, It returns only the preview URL and ID's. It uses these ID's to download it from `/mercury/attachments/video/?video_id=...`. This could be another feature I'll work on.

fbchat/_file.py
krzesu0: image
Response looks like this so I don't a lot of items to work with.
`image` is a thumbnail (no size given),
`image1` is a medium preview (size given, parsed by \_\_init\_\_),
`image2` is a original **(?)** file (size given, parsed by \_\_init\_\_).

fbchat/_client.py
krzesu0: Yup, changed it when i merged both function together.

fbchat/_client.py
krzesu0: Fixed on function merge.

fbchat/_file.py
krzesu0: The same is is for video btw.

fbchat/_client.py
krzesu0: Yup. Easy!

fbchat/_file.py
madsmtm: Whoops, I forgot about the parsing in `__init__` 👍

fbchat/_client.py
```diff
@@ -976,6 +976,41 @@ def getUserActiveStatus(self, user_id):
...
+        :param thread_id: ID of the thread
+        :return: :class:`ImageAttachment` or :class:`VideoAttachment`.
+        :rtype: iterable
+        """
+        thread_id, thread_type = self._getThread(thread_id, None)
+        data = {"id": thread_id, "first": 48}
+        j = self.graphql_request(_graphql.from_query_id("515216185516880", data))
+        while True:
+            try:
+                i = j[thread_id]["message_shared_media"]["edges"][0]
```
madsmtm: The ID is a string key in the response, so please make sure to convert the `thread_id` before using it here (so that the user can pass an `int`, and it'll still work fine)

fbchat/_file.py
```diff
@@ -155,6 +155,18 @@ def _from_graphql(cls, data):
...
+            preview=data["image1"],
+            uid=data["id"],
```
madsmtm: Use the attachment ID instead of the ID returned by the endpoint.

```suggestion
            uid=data["legacy_attachment_id"],
```

fbchat/_file.py
```diff
@@ -252,6 +264,18 @@ def _from_subattachment(cls, data):
...
+    @classmethod
+    def _from_list(cls, data):
+        data = data["node"]
+        return cls(
+            width=data["original_dimensions"].get("x"),
+            height=data["original_dimensions"].get("y"),
+            small_image=data["image"],
+            medium_image=data["image1"],
+            large_image=data["image2"],
+            uid=data["id"],
```
madsmtm: Use the attachment ID instead of the ID returned by the endpoint.

```suggestion
            uid=data["legacy_attachment_id"],
```

fbchat/_client.py
```diff
@@ -976,6 +976,41 @@ def getUserActiveStatus(self, user_id):
...
+                    continue
+                else:
+                    break
+
+            if i["node"].get("__typename") == "MessageImage":
+                yield ImageAttachment._from_list(i)
+            elif i["node"].get("__typename") == "MessageVideo":
+                yield VideoAttachment._from_list(i)
+            else:
+                yield Attachment(uid=i["node"]["uid"])
```
madsmtm: Use `legacy_attachment_id` instead of `uid`. And for some attachments, such as `MessageAudio` (which this endpoint also return), you won't even get an id!

```suggestion
                yield Attachment(uid=i["node"].get("legacy_attachment_id"))
```

fbchat/_client.py
krzesu0: > so that the user can pass an int, and it'll still work fine

Oh, yeah, good point.

PR #443, opened 2019-07-04
Fix 2FA login error

### LINKED ISSUE #442: 2FA login error
## Description of the problem
When 2FA is enabled, instead of prompting to enter input, an error appears and another attempt is done - the script is stuck at loop.

## Code to reproduce
```py
from fbchat import Client
client = Client('<email>', '<password>')
```

## Traceback
```
Logging in <email>...
Attempt #1 failed, retrying
Traceback (most recent call last):
  File "/home/u/.local/lib/python3.7/site-packages/fbchat/_client.py", line 254, in login
    state = State.login(email, password, user_agent=user_agent)
  File "/home/u/.local/lib/python3.7/site-packages/fbchat/_state.py", line 127, in login
    code = on_2fa_callback()
NameError: name 'on_2fa_callback' is not defined
```

## Environment information
- python 3.7.3
- fbchat 1.7.2 (installed via pip)

kapi2289:
Closes #442

### DIFF
--- a/fbchat/_client.py
+++ b/fbchat/_client.py
@@ -2925,10 +2925,6 @@ class Client(object):
         """
         log.info("Logging in {}...".format(email))
 
-    def on2FACode(self):
-        """Called when a 2FA code is needed to progress"""
-        return input("Please enter your 2FA code --> ")
-
     def onLoggedIn(self, email=None):
         """
         Called when the client is successfully logged in
--- a/fbchat/_state.py
+++ b/fbchat/_state.py
@@ -81,6 +81,11 @@ def _2fa_helper(session, code, r):
     return r
 
 
+def on_2fa_callback():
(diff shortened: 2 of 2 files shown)


fbchat/_client.py
```diff
@@ -2925,10 +2925,6 @@ def onLoggingIn(self, email=None):
         """
         log.info("Logging in {}...".format(email))
 
-    def on2FACode(self):
```
kapi2289: @madsmtm Should we remove this or rather leave it as it is? This function won't be called anymore, so I think we can just do this small breaking change.

madsmtm:
I decided to fix this a little differently, in a way that we don't break backwards compatibility, see e136d77ade761bf34de5d23e81d3605c8879ff07

PR #445, opened 2019-07-05
Added heart reaction

kugo12:


madsmtm:
I've verified that this works while listening for messages, when fetching messages and when reacting to messages.

Thanks 👍

nikatark:
how do i  install / add this to my messenger ? sorry new to github.

ekohilas:
@nikatark What are you trying to acomplish?
This module is meant to emulate what messenger does in the background, so it wouldn't make much sense to "install" it into messenger.

nikatark:
i wanted to get the heart emoji reaction , some people even have the + icon allowing them to use any emoji as a reaction so i was trying to get that.

madsmtm:
Hi there @nikatark. The functionality you're describing only works on mobile devices, as far as I can figure out, Facebook hasn't made a way that this can be done elsewhere.

Issue #446, opened 2019-07-06
login fails, but active sessions are created on fb account


sabrinashie:
## Description of the problem
Sorry about posting another login issue, but I can't seem to login even though the username and password are definitely correct, especially since I can see all 5 active sessions documented when I take a look at my FB profile. 

I am not sure what's causing "home" to not be part of the url after getting /login/save-device/cancel/. Traceback is below:

## Code to reproduce
```
#!/usr/bin/env python3

from fbchat import Client
import getpass
import logging
logging.basicConfig(level=logging.DEBUG)
client = Client("<email>", getpass.getpass(prompt="Password: "))
```
## Traceback
```
Logging in <email>...
INFO:client:Logging in <email>...
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): m.facebook.com:443
DEBUG:urllib3.connectionpool:https://m.facebook.com:443 "GET / HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:https://m.facebook.com:443 "POST /login.php?login_attempt=1 HTTP/1.1" 302 0
DEBUG:urllib3.connectionpool:https://m.facebook.com:443 "GET /login/save-device/?login_source=login&refsrc=https%3A%2F%2Fwww.facebook.com%2F&_rdr HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:https://m.facebook.com:443 "GET /login/save-device/cancel/ HTTP/1.1" 302 0
DEBUG:urllib3.connectionpool:https://m.facebook.com:443 "GET /?_rdr HTTP/1.1" 200 None
Attempt #1 failed, retrying
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/fbchat/_client.py", line 254, in login
    state = State.login(email, password, user_agent=user_agent)
  File "/usr/local/lib/python3.7/site-packages/fbchat/_state.py", line 139, in login
    "(Failed on url: {})".format(r.url)
fbchat._exception.FBchatUserError: Login failed. Check email/password. (Failed on url: https://m.facebook.com/?_rdr)
ERROR:client:Attempt #1 failed, retrying

```
And onward, 4 more times.
## Environment information
- Python version 3.7
- `fbchat` version 1.7.2

skrobchik:
I'm having the same issue. The password and email is correct, I quadruple checked ten times already. Whenever I try to login with fbchat it will throw FBchatUserError and I'll see a new logged in device in Facebook settings.

[...]

PR #449, opened 2019-07-17
Fix login error

GarryLai:
https://github.com/carpedm20/fbchat/issues/446


madsmtm:
I agree with @kapi2289, so I'll close this for now

PR #464, opened 2019-08-28
Move requests into state

madsmtm:
Move various request helpers (e.g. `Client._post`) into the `State` model. This is done so that we can slowly start moving implementations into the relevant models (instead of having them live in `Client`).

For example, the implementation for `Client.addUsersToGroup` could, and should, be moved to `Group.add_users`. ~(However, before we can do this, a bit more work is required on updating `._getSendData` and `._doSendRequest`)~

This PR should be fairly trivial to review, one commit at a time.

### DIFF
--- a/fbchat/_client.py
+++ b/fbchat/_client.py
@@ -1305,30 +1305,7 @@ class Client(object):
         )
 
     def _upload(self, files, voice_clip=False):
-        """Upload files to Facebook.
-
-        `files` should be a list of files that requests can upload, see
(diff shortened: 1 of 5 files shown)


madsmtm:
Update: I've merged `Client._getSendData` and `Client._doSendRequest`, and moved them to `State._do_send_request`. So we should be close to the point where the helper / utility methods in `Client` has been moved to `State` or removed!

This means we can (finally) start moving `Client` methods to the relevant classes

PR #471, opened 2019-09-07
Update _util.py

adbenitez:
`mimetypes.guess_type` fails if the url is something like: http://example.com/file.zip?u=10 
this fix it.


madsmtm:
Thanks for the PR, I've backported it to `v1` in bb1f7d929478942a07af20d9cce44418a9f9c7b8, and released `v1.8.3` 👍

PR #492, opened 2019-12-26
Add a optional timestamp parameter to mark_as_read and mark_as_unread

OneBlue:
Currently `mark_as_read()` and `mark_as_unread()` call now() to determine the read timestamp to push to Facebook.

This can create a race condition if:

* Caller read message m1
* Caller calls `mark_as_read()`
* In the meantime, another message, m2 is received in the same thread
* `_read_status()` gets its timestamp calling `now()`, which is after `m2.timestamp` , but m2 hasn't been read yet
* As a result, m2 is marked as read while it hasn't been

This change adds an optional `timestamp` argument to `mark_as_read()` and `mark_as_unread()` to solve this issue (the caller *should* set this argument to the timestamp of the last message that been read)

### DIFF
--- a/fbchat/_client.py
+++ b/fbchat/_client.py
@@ -1722,41 +1722,46 @@ class Client:
         j = self._payload_post("/ajax/mercury/delivery_receipts.php", data)
         return True
 
-    def _read_status(self, read, thread_ids):
+    def _read_status(self, read, thread_ids, timestamp=None):
         thread_ids = _util.require_list(thread_ids)
 
-        data = {"watermarkTimestamp": _util.now(), "shouldSendReadReceipt": "true"}
+        data = {
+            "watermarkTimestamp": timestamp or _util.now(),
+            "shouldSendReadReceipt": "true",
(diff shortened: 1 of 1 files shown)


fbchat/_client.py
```diff
@@ -1722,41 +1722,46 @@ def mark_as_delivered(self, thread_id, message_id):
...
 
-    def mark_as_unread(self, thread_ids=None):
+    def mark_as_unread(self, thread_ids=None, timestamp=None):
         """Mark threads as unread.
 
         All messages inside the specified threads will be marked as unread.
 
         Args:
             thread_ids: User/Group IDs to set as unread. See :ref:`intro_threads`
+            timestamp: Timestamp to signal the read cursor at, in milliseconds, default is now()
```
madsmtm: Perhaps this should be worded as "default is the current time" instead of "default is now()"

OneBlue:
Thanks for the review @madsmtm. I have updated the branch.

PR #559, opened 2020-04-28
First Object then File Pointer

xaadu:
json.dump() receives object as first argument and File Pointer as 2nd argument.


madsmtm:
Thanks 👍

xaadu:
It's a pleasure!

PR #562, opened 2020-05-08
Update fetch.py

smilexs4:
Solved the exception:
`client = fbchat.Client(session)
TypeError: __init__() takes 1 positional argument but 2 were given`

By the way I can't find the following function. Could it be missing?
`client.fetch_user_info()`


madsmtm:
I removed `fetchUserInfo` since it was basically `fetch_thread_info`, but with a check to verify that you only fetched users. Seemed redundant in my eyes.

Thanks for the fix though 👍

Issue #584, opened 2020-05-29
Problem Logging In


benvchurch:
When I run the sample code:

from fbchat import Client
import logging
client = Client('<email>', '<pw>', logging_level=logging.DEBUG)

I get the following error (replicated on two different machines)

File "/usr/local/lib/python2.7/dist-packages/fbchat/_client.py", line 103, in __init__
    self.login(email, password, max_tries, user_agent=user_agent)
 File "/usr/local/lib/python2.7/dist-packages/fbchat/_client.py", line 209, in login
    user_agent=user_agent,
  File "/usr/local/lib/python2.7/dist-packages/fbchat/_state.py", line 149, in login
    return cls.from_session(session=session)
  File "/usr/local/lib/python2.7/dist-packages/fbchat/_state.py", line 186, in from_session
    fb_dtsg = FB_DTSG_REGEX.search(r.text).group(1)
AttributeError: 'NoneType' object has no attribute 'group'

p
I am running the most recent version of fbchat in python 2.7.17.

wetmore:
Have you tried logging in manually to the account you are using? Perhaps the account has been locked?

benvchurch:
Yes, there is no such issue.

qwertyuu:
Yup the problem is on Facebook's side!

Issue #556, opened 2020-04-19
Error with logout function


Callen93:
## Description of the problem
When trying to logout an active session created from session cookies, an error occurs. 

## Code to reproduce
```
session = fbchat.Session.from_cookies(cookie)
client = fbchat.Client(session=session)
session.logout()  
```

## Traceback
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rmt/Documents/fbchat-master/fbchat/_session.py", line 289, in logout
    logout_h = re.search(r'name=\\"h\\" value=\\"(.*?)\\"', h_r.text).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

## Environment information
- Python3.5
- `fbchat` 2.0.0a2

This error is not exclusive to v2. I'm getting same error using version 1.8.1 of fbchat. However I don't know if this error already existed or if a modification of fb handling of logout broke something (I've never used this functionality).

Issue #567, opened 2020-05-10
Cannot login 


kdalanon-ninthwave:
Logging fails with specific user. I am not using client.logout() in my code 
The program works fine with other facebook accounts but it doesn't work on this one.

## Code to reproduce

def getCredentials():

	#Retrieves the values
	with open('credentials.txt') as json_file:
		data = json.load(json_file)
		for p in data['credentials']:
			email = p['email']
			password = p['password']
			message = p['message']
	return email,password,message

credentials = getCredentials()
message = credentials[2]
client = Client(credentials[0], credentials[1])

## Traceback
```
Attempt #4 failed, retrying
Traceback (most recent call last):
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_client.py", line 205, in login
    self._state = State.login(
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_state.py", line 149, in login
    return cls.from_session(session=session)
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_state.py", line 186, in from_session
    fb_dtsg = FB_DTSG_REGEX.search(r.text).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
Traceback (most recent call last):
  File "fb.py", line 357, in <module>
    client = Client(credentials[0], credentials[1])
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_client.py", line 103, in __init__
    self.login(email, password, max_tries, user_agent=user_agent)
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_client.py", line 205, in login
    self._state = State.login(
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_state.py", line 149, in login
    return cls.from_session(session=session)
  File "C:\Users\kdalanon\AppData\Local\Programs\Python\Python38\lib\site-packages\fbchat\_state.py", line 186, in from_session
    fb_dtsg = FB_DTSG_REGEX.search(r.text).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

## Environment information
- Python version 3
- `fbchat` version
- If relevant, output from `$ python -m pip list`

If you have done any research, include that.
Make sure to redact all personal information.

[...]

Issue #555, opened 2020-04-18
Login failed for 2FA enabled account.


MS-Jahan:
## Description of the problem
Can't login to the account which have 2FA enabled. Login works fine with the account which doesn't have 2FA enabled.

## Code I'm using:

```
def login_logout():
    cookies = {}
    try:
        # Load the session cookies
        if os.path.isfile('session.json'):
            with open('session.json', 'r') as f:
                cookies = json.load(f)
    except:
        os.remove('session.json')
        # If it fails, never mind, we'll just login again
        # client = CustomClient(email, password, max_tries=1)
    if((not cookies) != True):
        client = CustomClient(email, password, session_cookies=cookies, user_agent=user_agent, max_tries=1)
    else:
        client = CustomClient(email, password, user_agent=user_agent, max_tries=1)
    
    with open('session.json', 'w') as f:
        json.dump(client.getSession(), f)

    print("\nProgram Started!\n")
    client.listen()
```

# Error I'm getting

```
Traceback (most recent call last):
  File "C:\Users\USERNAME\AppData\Local\Programs\Python\Python37\lib\site-pa
ckages\fbchat\_client.py", line 175, in setSession
    self._state = State.from_cookies(session_cookies, user_agent=user_agent)
  File "C:\Users\USERNAME\AppData\Local\Programs\Python\Python37\lib\site-pa
ckages\fbchat\_state.py", line 208, in from_cookies
    return cls.from_session(session=session)
  File "C:\Users\USERNAME\AppData\Local\Programs\Python\Python37\lib\site-pa
ckages\fbchat\_state.py", line 186, in from_session
    fb_dtsg = FB_DTSG_REGEX.search(r.text).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

## Environment information
- Python version 3.6.8
- `fbchat` version 1.9.6

PR #586, opened 2020-06-01
Fix AttributeError on login

qwertyuu:
Facebook decided they needed an "Accept" header to return data on request and that broke fbchat

### DIFF
--- a/fbchat/_state.py
+++ b/fbchat/_state.py
@@ -27,6 +27,8 @@ def find_input_fields(html):
 def session_factory(user_agent=None):
     session = requests.session()
     session.headers["Referer"] = "https://www.facebook.com"
+    session.headers["Accept"] = "text/html"
+
     # TODO: Deprecate setting the user agent manually
     session.headers["User-Agent"] = user_agent or random.choice(_util.USER_AGENTS)
     return session


qwertyuu:
Seems to fix:
#584 
#567 
#555
#551 
#565 

Maybe fixes
#556

Callen93:
Hello,
I don't know if it's the right place to discuss this but as it seems to exists multiple login/logout problem right now I want to expose some conclusion and try to find solution. Thank you by the way for this fix.

I think the update of facebook with new UI did broke a lot of things. The thing is activating/desactivating the UI change (or at least changed) the behaviour of facebook against fbchat. For instance on friday (can't reproduce today), I could login with fbchat only when I desactivated new UI. So one fix can work for people with old UI and not people with new UI.

Lets first focus on branch  carpedm20:v1 :
- Connecting with session cookies work with or without your fix. 
- Trying to connect with login password fail and block account redirecting causing this error  `fbchat._exception.FBchatUserError: Login failed. Check email/password. (Failed on url: https://m.facebook.com/checkpoint/?_rdr) `. This is even the case when using the same user agent as when, connecting with webUI. https://github.com/carpedm20/fbchat/issues/477 https://github.com/carpedm20/fbchat/issues/532 
- I can't reproduce the error  `'NoneType' object has no attribute 'group'` https://github.com/carpedm20/fbchat/issues/567 https://github.com/carpedm20/fbchat/issues/555 https://github.com/carpedm20/fbchat/issues/551 https://github.com/carpedm20/fbchat/issues/565  https://github.com/carpedm20/fbchat/issues/584 
- https://github.com/carpedm20/fbchat/issues/556 is not fixed by this commit 

On master version, it's a bit worse. I used your change on master version and get 
- When trying to connect with session cookies `fbchat.NotLoggedIn: Found empty fb_dtsg, the session was probably invalid.`
- If trying to apply proposed changes in https://github.com/carpedm20/fbchat/issues/552 same problem but message error is `fbchat.NotLoggedIn: Could not find fb_dtsg`
- When trying to connect via login/password, It still fail and block temporary my account.

I'm still trying to investigate on this and come back at you if I find something interesting.

qwertyuu:
@Callen93 Things I can tell you about my setup and why it fixed the problems I had:

1. I am *not* using the new Facebook UI
1. I do *not* have 2FA (just stating this makes me want to enable it, I feel like someone will scan through the github comments and try to hack me)

I went through the real `messenger.com` and saw that the headers differed from what the "Sessions" client headers tried to send to facebook. By replicating my browser's headers, I was able to get a response in `brotli` form. When I commented out all the unnecessary headers, I ended up with only the one I uploaded through the fix. My initial problem was `'NoneType' object has no attribute 'group'` and it happened because `r.text` at that place was an empty string (facebook just didn't respond any content but it as a 200 OK response anyways). You are right, I think all these login/logout problems are from the new UI update and may have impacted the old one too.

madsmtm:
Hmm, we're using `requests`, which should set the `Accept: */*` pr. default, which is the same thing browsers do! So I don't really understand this, are you _sure_ this actually fixes anything, and isn't just by coincidence that it suddenly worked?

Regarding the new UI, that is partly the reason why I changed the login logic on the `master` branch to use messenger.com urls, see 079d4093c4f0febf004c6cf8f6972f157833a4f7. Haven't had the time to backport a lot of the fixes (and bugs) in there to `v1`, so that's the primary reason `v1` can't handle the new UI

madsmtm:
Theory: Using `Accept: text/html` triggers legacy code, so that the new UI isn't activated?

qwertyuu:
> are you _sure_ this actually fixes anything, and isn't just by coincidence that it suddenly worked?

This is a hard question because I don't have time to test it elsewhere and it's hard to be _sure_ unless many others try it and report back as positive. In my case it was a causal effect: adding that line made fbchat login, while removing it had a devastating effect (empty response body so all the regexes fired on an empty string)

The only way I can be sure it wors is that @Callen93 tried it and reported back as positive, and this is the closest I can be to being _sure_ it at least fixes the login procedure.

madsmtm:
Well, I think that's good enough for me, I'll take it, thanks!

